PR #48, opened 2024-01-24
Find this plugin by its currrent name "CSharpForUE" instead of "UnrealSharp"

cronofear-dev:


OlssonDev:
Use UE_PLUGIN_NAME instead. So if we end up renaming the plugin, this won't break

PR #49, opened 2024-01-24
Add a setting to require editor focus for Hot Reload

tzainten:
I tend to hit CTRL + S very frequently when I'm typing code, so I added a small toggle to hold off on Hot Reloading until Unreal Engine regains focus


OlssonDev:
Looks great! Thank you for the pull request! :D

PR #50, opened 2024-01-26
Fix changes not being loaded on engine startup

tzainten:
I've noticed that any changes I make to my managed assembly only get loaded whenever I save a file while the editor is open. 

So I hooked `FCoreDelegates::OnAllModuleLoadingPhasesComplete` to call `FUnrealSharpEditorModule::Reload` when the engine finishes loading all modules.

Keep in mind that if the managed assembly contains any errors, the engine will not finish starting up and will continuously prompt you until that error gets fixed. I'm unsure if this is a good thing, but these changes seem to work for me.

### DIFF
--- a/Managed/UnrealSharpPrograms/UnrealSharpBuildTool/Actions/GenerateProject.cs
+++ b/Managed/UnrealSharpPrograms/UnrealSharpBuildTool/Actions/GenerateProject.cs
@@ -98,6 +98,12 @@ public class GenerateProject : BuildToolAction
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


OlssonDev:
This can be fixed in the UnrealSharpBuildTool during the GenerateProject action.

Currently it GeneratesProject => Build DLL. Add the step of weaving after that and it should have the same result.

tzainten:
Is this what you meant? This also seems to work for me

OlssonDev:
Yes! :D 

I'll check it out tomorrow if it works on my end too.

PR #61, opened 2024-02-06
Fix UFunctions not applying the metadata defined in C# and Fix some PropertyFlags mismatch with how they work C++

cronofear-dev:
First commit fixes an issue where metadatas weren't being applied when used in UFunctions: 
```csharp
// This works correctly
[UProperty(PropertyFlags.EditAnywhere | PropertyFlags.BlueprintReadWrite), UMetaData("DisplayName","MyInteger"), UMetaData("Category","Inner|TestCat")]
public int TestInt { get; set; }
    
// Both UMetaData weren't being applied
[UFunction(FunctionFlags.BlueprintCallable), UMetaData("DisplayName", "MyTestFunction"), UMetaData("Category","Inner|TestCat")]
public void TestFunction()
{
}
```

Second commit makes some PropertyFlags work closer to how they work in C++ (I haven't tried every combination, but so far, I haven't encountered an issue). Also fixes the following issue where a combination of BlueprintReadWrite and EditAnywhere was causing the property to be ReadOnly in BP:

```csharp
// This combination caused TestInt to be readonly in BP
[UProperty(PropertyFlags.BlueprintReadWrite | PropertyFlags.EditAnywhere)]
public int TestInt { get; set; }
```

Sorry I'm not too familiar with github PRs, a third commit was auto-added to this PR. This one is about supporting the `DisplayName` and `Category` metadata directly in the UProperty and UFunction attributes:

```csharp
// So we can do this
[UProperty(PropertyFlags.BlueprintReadWrite | PropertyFlags.EditAnywhere, DisplayName = "MyNewName", Category = "Inner|TestCat")]
public string TestString { get; set; }

// Instead of this (which is still supported)
[UProperty(PropertyFlags.BlueprintReadWrite | PropertyFlags.EditAnywhere), UMetaData("DisplayName", "MyNewName"), UMetaData("Category","Inner|TestCat")]
public string TestString { get; set; }
```

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/Attributes/UFunctionAttribute.cs
+++ b/Managed/UnrealSharp/UnrealSharp/Attributes/UFunctionAttribute.cs
@@ -42,7 +42,7 @@ public enum FunctionFlags
 
 [AttributeUsage(AttributeTargets.Method)]
 [FunctionFlagsMap(NativeFunctionFlags.Native)] 
-public sealed class UFunctionAttribute(FunctionFlags flags = FunctionFlags.None) : Attribute
+public sealed class UFunctionAttribute(FunctionFlags flags = FunctionFlags.None) : BaseUAttribute
 {
     public FunctionFlags Flags { get; private set; } = flags;
 }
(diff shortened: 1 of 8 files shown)


OlssonDev (approved):
Thank you for the contribution! ❤️

PR #87, opened 2024-03-25
Added proper handling of exceptions thrown by managed events

thomasfn:
Currently exceptions thrown in a C# UFunction are swallowed up and logged to somewhere unuseful. This PR changes the managed method invoke mechanism to allow exceptions to be passed back to C++ and injected into the stack frame, allowing the BP error handler to take over. Most importantly this shows the exception in the output log as it occurs, which makes debugging considerably easier.

Example code:
```C#
public void TestException()
{
    throw new Exception("Test exception!");
}
```

image

The errors are marked as fatal which seems to end the PIE session immediately - I'm not sure if that's too severe and maybe they should only be warnings instead? Then again exceptions are pretty bad and the user can always wrap their own code in try/catch so maybe this is ok.

### DIFF
--- a/Source/CSharpForUE/TypeGenerator/CSClass.cpp
+++ b/Source/CSharpForUE/TypeGenerator/CSClass.cpp
@@ -7,6 +7,8 @@ void UCSClass::InvokeManagedMethod(UObject* ObjectToInvokeOn, FFrame& Stack, RES
 {
 	const UCSFunction* Function = CastChecked<UCSFunction>(Stack.CurrentNativeFunction);
 	TArray<uint8> ArgumentData;
+	FString ExceptionMessage;
+	bool Success;
 
 	// Skip allocating memory for the argument data if there are no parameters that need to be passed
 	if (Function->NumParms == 0)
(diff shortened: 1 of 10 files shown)


Source/CSharpForUE/TypeGenerator/CSClass.cpp
```diff
@@ -95,8 +108,18 @@ void UCSClass::ProcessOutParameters(FOutParmRec* OutParameters, uint8* ArgumentD
...
 	const FGCHandle ManagedObjectHandle = FCSManager::Get().FindManagedObject(ObjectToInvokeOn);
-	FCSManagedCallbacks::ManagedCallbacks.InvokeManagedMethod(ManagedObjectHandle.GetHandle(), Function->GetManagedMethod(), ArgumentData.GetData(), RESULT_PARAM);
+	struct { TCHAR* Data = nullptr; int ArrayNum = 0; int ArrayMax = 0; } ExceptionText;
+	int ResultCode = FCSManagedCallbacks::ManagedCallbacks.InvokeManagedMethod(ManagedObjectHandle.GetHandle(), Function->GetManagedMethod(), ArgumentData.GetData(), RESULT_PARAM, &ExceptionText);
+	if (ResultCode != 0)
+	{
+		ExceptionMessage = FString(static_cast<TCHAR*>(ExceptionText.ArrayNum, ExceptionText.Data));
+		// Commented as causing crash - possibly due to StringMarshaller allocating on the COM heap rather than the normal heap?
+		// Do we need a better way of passing strings to unreal without causing memory leaks?
+		// FMemory::Free(ExceptionText.Data);
```
thomasfn, line 121: As comment says this needs some thinking about. I'm not sure why the COM heap is being used instead of the normal heap, and I'm not sure how the StringMarshaller (and ArrayMarshaller) avoids memory leaks. Ideally we'd let C# free the memory since it's the one allocating it, but I'm not sure how to let it know when then memory is ready to be free'd without adding another interop call here.

Issue #76, opened 2024-02-17
Support for async actions in C#


CadeEvs:
The ability for the plugin and generate and support blueprint async actions

rob-ack:
@OlssonDev This issue can be closed as async actions are supported by #91 which is merged

PR #91, opened 2024-03-28
Add support for async actions

thomasfn:
Addresses https://github.com/UnrealSharp/UnrealSharp/issues/76.

This adds `CSBlueprintAsyncActionBase` that can be extended in C#, much like the native `UBlueprintAsyncActionBase` that can be extended in C++. This allows user code in C# to create latent actions which are useful for asynchronous workflows in blueprint. This also adds glue code generation for native actions that derive `UBlueprintAsyncActionBase` so that they can be used from C#.

Also fixed in this PR:
- `NewObject` crashing if you don't pass an explicit `SubclassOf` (as the default parameter gives you a struct full of nulls essentially)
- Support for parameters in delegates (both native and user-defined)

Example code and usage:
```C#
public partial class OnCompletedDelegate : MulticastDelegate<OnCompletedDelegate.Signature>
{
    public delegate void Signature(int Result);
}

[UClass]
public class TestLatentMethodAsyncAction : CSBlueprintAsyncActionBase
{
    [UProperty(PropertyFlags.BlueprintAssignable)]
    public OnCompletedDelegate OnCompleted { get; set; }

    [UProperty]
    public int LhsValue { get; set; }

    [UProperty]
    public int RhsValue { get; set; }

    protected override void ReceiveActivate()
    {
        OnCompleted.Invoke(LhsValue + RhsValue);
    }

    [UFunction(FunctionFlags.BlueprintCallable), UMetaData("BlueprintInternalUseOnly", "true"), UMetaData("Category", "Test Async"), UMetaData("DisplayName", "Test Latent Method (Async)")]
    public static TestLatentMethodAsyncAction TestLatentMethod(UnrealSharp.CoreUObject.Object outer, int lhsValue, int rhsValue)
    {
        var action = NewObject<TestLatentMethodAsyncAction>(outer);
        action.LhsValue = lhsValue;
        action.RhsValue = rhsValue;
        return action;
    }
}
```
image

```C#
[UClass]
public class TestSceneCaptureActor : Actor
{
    [UProperty]
    private AsyncCaptureScene? AsyncCaptureSceneAction { get; set; }

    [UProperty(DefaultComponent = true, RootComponent = true)]
    private SceneComponent Root { get; set; }

    [UProperty(DefaultComponent = true)]
    private CameraComponent Camera { get; set; }


    public void TestCaptureScene()
    {
        AsyncCaptureSceneAction = AsyncCaptureScene.CaptureSceneAsync(Camera, new SubclassOf<SceneCapture2D>(), 1024, 1024);
        AsyncCaptureSceneAction.Complete.Add(OnCaptureSceneComplete);
        AsyncCaptureSceneAction.Activate();
        PrintString("Started capturing scene...");
    }

    [UFunction]
    private void OnCaptureSceneComplete()
    {
        AsyncCaptureSceneAction = null;
        PrintString("Scene capture complete!");
    }
}
```

Unfortunately I did have to make our own version of `UK2Node_AsyncAction` (and put it in it's own module to make unreal happy), as the default unreal one was getting confused about the dummy blueprint that is associated with C# generated UClass's and kept trying to associate the latent nodes with those, but since they aren't real they just wouldn't show up as an option. Luckily most of the meat of the node is contained in a base class which I was able to use.


Source/UnrealSharpBlueprint/K2Node_CSAsyncAction.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include "K2Node_CSAsyncAction.generated.h"
+
+class FBlueprintActionDatabaseRegistrar;
+class UObject;
+
+UCLASS()
+class UK2Node_CSAsyncAction : public UK2Node_BaseAsyncTask
+{
+	GENERATED_UCLASS_BODY()
```
OlssonDev, line 18: GENERATED_BODY() instead

OlssonDev (changes requested):
What's the difference between sparse and inline delegate? I just copied other language integrations on that matter.

PR #108, opened 2024-04-09
Update README.md code snippet with syntax-highlighting

not-nugget:
I noticed that the code snippet in the README was not syntax highlighted, so I went ahead and updated it here. I also cloned the Wiki and updated all of the snippets and code blocks there with the correct language semantics so they would be highlighted, as well as made a few style changes. I cannot find a way to make a PR for a repository's wiki, however, but you can see the changes I made in my [forked repository](https://github.com/not-nugget/UnrealSharp/wiki).


OlssonDev:
Thank you! 😄

PR #109, opened 2024-04-10
Fixed memory leak when using Name.ToString.

thomasfn:
As far as I can tell this is the only case where a string was being leaked. I haven't changed arrays though and it's possible they need similar treatment.


OlssonDev:
Good catch! :D Thank you!

PR #110, opened 2024-04-11
fix duplicate root components

Rynmiles:
fix duplicate root components in blueprint editor


OlssonDev:
Thank you for the PR :D

PR #116, opened 2024-04-14
fix InvokeManagedMethod crash.

Rynmiles:
the function must have FUNC_HasOutParms flag with any properties have CPF_OutParm flag


OlssonDev:
Thank you for the PR! :D

PR #117, opened 2024-04-14
Fix requireFocusForHotReload option (needs `bIsReloading = true;` bef…

cronofear-dev:
…ore to work)


OlssonDev:
Awesome! Thank you :D

PR #130, opened 2024-04-26
Fixed includes, TEXT macros, linking and cast issues

MrRobinOfficial:
I tried this plugin using Windows 11 and Unreal Engine 5.4. Got a lot of compiling issues. Mostly it was missing includes and forgotten TEXT macros.

I also switch from using `Cast` to `CastField` inside (Source/GlueGenerator/CSharpGeneratorUtilities.cpp:150)

I also changed 'RF_PendingKill' to 'RF_MirroredGarbage' inside (Source/CSharpForUE/Extensions/CSCancellableAsyncAction.cpp:10)

Committed by Robin Johannesson.

### DIFF
--- a/Source/CSharpForUE/CSManager.cpp
+++ b/Source/CSharpForUE/CSManager.cpp
@@ -14,6 +14,7 @@
 #include "Misc/MessageDialog.h"
 #include "Engine/Blueprint.h"
 #include "UnrealSharpProcHelper/CSProcHelper.h"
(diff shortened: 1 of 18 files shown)


OlssonDev:
Awesome! Thank you for the PR!

I want to keep 5.3 compiling for a while, since 5.4 just released. There are some macros and includes that don't exist in 5.3 that exists in 5.4.

PR #147, opened 2024-05-27
Fixes issues with arrays in delegates, events and rpcs

FailMelon:
Arrays were currently broken due to a bunch of IL issues, this fixes a bunch of IL order mistakes as well as other some other stuff. Also makes them play nice with blueprints as well. Wasn't happy with where i put the check for ParameterType modifier but it seemed to align better with the current code, maybe could have the NativeDataType somehow access to override flags.


FailMelon:
Had some issues with it being an ref/out param weaver side so i made it into a serialized json value instead that then applies the correct flags unreal side.

OlssonDev:
Thank you for all the pull requests! :D I'll look through them soon ❤️

OlssonDev:
Thanks again for the PR! Nice one :D

I'll look at the other PRs the coming days.

PR #148, opened 2024-05-27
Added helper methods to Tasks for async context

FailMelon:
Some useful helper methods for async.

Instead of

```[UFunction(FunctionFlags.BlueprintCallable), UMetaData("DisplayName", "Slow Add (Async)")]
public async Task<int> SlowAdd(int lhs, int rhs, CancellationToken cancellationToken)
{
    SynchronizationContext.SetSynchronizationContext(UnrealSynchronizationContext.GetContext(NamedThread.GameThread));

    PrintString($"Commencing the world's slowest addition...");

    await Task.Delay(1000, cancellationToken);
    if (cancellationToken.IsCancellationRequested || IsDestroyed) { return default; }

    int result = lhs + rhs;
    PrintString($"{lhs} + {rhs} = {result}!");

    return result;
}
```

Can now just simply do

```[UFunction(FunctionFlags.BlueprintCallable), UMetaData("DisplayName", "Slow Add (Async)")]
public async Task<int> SlowAdd(int lhs, int rhs, CancellationToken cancellationToken)
{
    PrintString($"Commencing the world's slowest addition...");

    await Task.Delay(1000, cancellationToken).ConfigureWithUnrealContext();
    if (cancellationToken.IsCancellationRequested || IsDestroyed) { return default; }

    int result = lhs + rhs;
    PrintString($"{lhs} + {rhs} = {result}!");

    return result;
}
```

This is useful because the context needs to be set after each task that can potentially yield to a different thread with the current usage, thus reducing the amount of bloat.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/UnrealSynchronizationContext.cs
+++ b/Managed/UnrealSharp/UnrealSharp/UnrealSynchronizationContext.cs
@@ -59,7 +59,87 @@ namespace UnrealSharp
         AnyBackgroundHiPriTask = AnyThread | BackgroundThreadPriority | HighTaskPriority,
     };
 
-    public class UnrealSynchronizationContext : SynchronizationContext
+    public static class UnrealContextTaskExtension
+    {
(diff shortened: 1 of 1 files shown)


OlssonDev:
Thank you for the PR! Awesome stuff!

PR #149, opened 2024-05-27
Fixes user assembly resolving managed dlls

FailMelon:
Currently if you use any kind of type in a ufunction or uproperty of such from another project it'll cause a weaver file lock. This fixes that by applying the same resolver the managed UnrealSharp project does before hand.

Let me know if you'd like me to also fix unmanaged dlls in this pr and will also work on that.


OlssonDev:
Sorry for the slow review. Been pretty busy fixing some serious issues with the plugin and packaging. 

What do you mean with unmanaged dlls in this context?

OlssonDev (approved):

OlssonDev:
Thank you for the pull request! I forgot to merge this last week when I looked into it :D

Issue #134, opened 2024-04-29
Mark non-API methods and types as internal


thomasfn:
For example, all the generated `FromNative` and `ToNative` glue code, the marshallers, `ReceiveX_Implementation` and other things that don't really need to be exposed to the user project. These could be marked `internal` instead of `public` to reduce how much "things you shouldn't directly use, call or override" are showing up in intellisense, to keep the api nice and tidy.

PR #150, opened 2024-05-28
Cleanup API surface

FailMelon:
Bunch of stuff that can be confusing to the end user should not be getting exposed such as marshallers and methods with _Implementation, this PR was made to cleanup the api surface to help guide users to use the correct api.

- [x] Make Marshallers internal
- [x] Make _Implementation methods internal
- [x] Make Exporters internal
- [x] Make UnrealSharp load from user binaries
- [x] Expose internals to their previous state.

Let me know if theres anything i've missed that should not be exposed to the end user.

#134


FailMelon:
With making things internal the user assembly will need to be friendly with UnrealSharp plugin via InternalsVisibleTo attribute, i not exactly sure what the best approach would be for this. Modifying the UnrealSharp assembly with Mono.Cecil during assembly load time i think is possibly out of the question because then the published build would require Mono.Cecil. Modifying the assembly during weaver time also doesn't currently seem like a possibility because of how UnrealSharp is loaded.

FailMelon:
After looking into this a bit more have decided the best approach would be to load UnrealSharp from the user binaries directory, UnrealSharp will then get weaved as a first time setup of the editor loading which exposes the neccessary internals to their previously respective access attributes.

OlssonDev:
I'll check the whole thing this coming week! :D After a quick glance, can we move the setup weaver to UHT?

Sorry for the delay.

PR #151, opened 2024-06-14
Add c# class extenion feature, fix bind failed due to variable namespace missing, add Compile button to editor

ySion:
1. Fixed the issue where launching a standalone window game from the editor would still trigger rebinding.
2. Fixed variable namespace missing (when you try to enable PCG plugin, variable namespace missing might cause bind failed)
3. Disable Some warnings in c# generation building, I think it will more easy to find errors in generate message.
### **Feature**

In the above example, I extend class **"Engine.Actor"** and **"Engine.Character"**
> **The first argument** need named as "myself", then UnrealSharp will replace it correctly when binding generating.
>
> **B. Compile it, use it in c# !**
>> image
>> image
>> image
>> It makes something easy:
>> image

**Warning**, **DO NOT WRITE C++ HELPER CLASS IN EDITOR ONLY PLUGIN! NOTICE IT!**
> **Don't write c++ helper class in any editor only plugin, it will works when running game in editor, but will crash in standalone mode and publish game without any logs.** 
> **But if you know what are you doing, skip it.**


OlssonDev:
Hi, thank you for the pull request! Cool stuff :D

Some general notes:
Not sure yet how I feel about the toggle button, but I like the compile button! :D
Do PascalCase instead of camelCase
Open/Close bracketon new line

I want the source to be close as source as possible, so it's easier to jump between the source code and the plugin

ySion:
Modifed 👍

OlssonDev:
Sorry for the slow review. 

I'm not too keen on the toggle button still! Other than that it looks good! :D

PR #160, opened 2024-07-19
Fix EngineDir not being set when compiling UnrealSharpScriptGenerator

tzainten:
This should fix the error `The imported project "C:\Source\Programs\Shared\UnrealEngine.csproj.props" was not found.`


OlssonDev:
Thank you for the PR :D

PR #169, opened 2024-07-26
Generate ManagedProject.csproj at compile time

tzainten:
I noticed that the ManagedProject.csproj file doesn't get generated until you boot up Unreal Engine. This change should generate it at compile time.

### DIFF
--- a/Source/CSharpForUE/CSManager.cpp
+++ b/Source/CSharpForUE/CSManager.cpp
@@ -66,7 +66,7 @@ void FCSManager::InitializeUnrealSharp()
 	{
 		if (!FApp::IsUnattended())
 		{
-			if (!FCSProcHelper::GenerateProject())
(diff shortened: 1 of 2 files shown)


OlssonDev:
Awesome :D

Maybe remove the GenerateProject in C++ then and just let it compile the user code?

tzainten:
I assume you're talking about making this change in _FCSManager::InitializeUnrealSharp()_:
```
#if WITH_EDITOR
	if (!FParse::Param(FCommandLine::Get(), TEXT("game"))) 
	{
		if (!FApp::IsUnattended())
		{
			/*
			if (!FCSProcHelper::GenerateProject())
			{
				InitializeUnrealSharp();
				return;
			}
			*/

			FCSProcHelper::InvokeUnrealSharpBuildTool(Build);
			FCSProcHelper::InvokeUnrealSharpBuildTool(Weave);
		}
	}
#endif
```

The above code does work for me in C++ projects. However, I noticed that in Blueprint Only projects this doesn't generate a project file.

PR #172, opened 2024-08-04
Expose public LoadPlugin and UnloadPlugin functions

CadeEvs:
Modifies the UnrealSharp.Plugins project to expose public functions for loading and unloading assemblies.

Only for advanced usage but very helpful if someone wants to leverage UnrealSharp's dll loading/unloading system.


OlssonDev:
Thank you for the PR :D

PR #173, opened 2024-08-06
Fix using UStruct in delegates

Netlex:


OlssonDev:
Thank you for the fix :D

PR #177, opened 2024-08-17
fix #176 -user breakpoints will not hit for user code for unreal event implementations

### LINKED ISSUE #176: [Debugging] Breakpoints for Actor Event (like begin play) implementations are never hit by debugger.
The reason for this is that during IL weaving the user code for these methods is copied into a new functions on IL level and all calls for are performed on these methods behind the scene.

This in turn makes the debugging fail to hit the user breakpoints since these match to an actual different code location from the code location that actually is called.

All this is hidden from the user.

rob-ack:
- fix user breakponits not hitting by copying over the metadata tokens of the original method.
- disabling pdb generation since it does not make sense will rewrite all debugging metadata wich prevents from matching code  IL code to user code manually (like with metadata tokens)

this will lead to matching debug symbols for IL method copies and breakpoints will hit like desired for user code.

i made more extensive tests with 
- calling event functions from blueprints
- implementing delegates and callbacks to regular ufunctions and event functions
- general [UFunction(FunctionFlags.BlueprintEvent | FunctionFlags.BlueprintCallable)] decorations

### DIFF
--- a/Managed/UnrealSharpPrograms/UnrealSharpWeaver/Program.cs
+++ b/Managed/UnrealSharpPrograms/UnrealSharpWeaver/Program.cs
@@ -127,7 +127,7 @@ public static class Program
         {
             assembly.Write(assemblyOutputPath, new WriterParameters
             {
-                WriteSymbols = true,
+                WriteSymbols = false,
                 SymbolWriterProvider = new PdbWriterProvider(),
             });
         }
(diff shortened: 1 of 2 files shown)


rob-ack:
image

OlssonDev (approved):
Awesome work! ⭐ Thank you for the PR! :D

PR #179, opened 2024-08-18
fix #178 -fix missing IL code for 'virtual' base method after weaving 

### LINKED ISSUE #178: [Crash | IL | Delegates] IL Code for 'virtual' BlueprintCallable base methods is not generated
image

The Il code for 'virtual' marked methods to invoke them via delegated or blueprints is missing.

If they are used from within a delegate U# will try to call it via its invoke system but the actual c# implementation will not be present and thus not be resolved by the underlying code. A null pointer is returned and the following U# code will grab into nothing making the engine crash.

rob-ack:
image

all virtual methods have base methods. change the code to check if the base method is not the current implementation. this base method should always be the root base method as well from the Cecil underlying impl


OlssonDev:
Awesome! ⭐ Thank you for the PR! :D

PR #180, opened 2024-08-18
fix #156 weaving failes on active debugger attached due to pdb files beeing locked 

### LINKED ISSUE #156: Can't hot reload while having debugger attached to managed code

rob-ack:
this will use a common move before write strategy to avoid file locks. a temporary directory is used and cleaned whenever possible.


OlssonDev:
Awesome fix! I was gonna look into this when I came home from my vacation ⭐

PR #182, opened 2024-08-22
Fixed SpawnActor transform

FailMelon:
This fixes an issue with SpawnActor transform sending FTransform rather then using the autogenerated marshalling


rob-ack (approved):
looks correct to me

FailMelon:
After talking with Olsson FTransform should be bittable but data is misaligned due to Translation and Scale3d being vector 4 double on the engine side.

PR #192, opened 2024-09-14
Fixed issue with weaver project paths

FailMelon:
Fixes a current issue with path in the BuildTool causing assembly to not be loaded in the assembly resolver


OlssonDev:
Thank you for the fix ⭐

PR #196, opened 2024-09-14
Fixed obj/bin folders not being ignored in Script subfolders

CadeEvs:
Someone was running into a problem where their other obj/bin folders weren't being ignored with the directory watcher, so now its checks for the folders without the Script prefix to account for subfolders


OlssonDev:
Awesome! Forgot about that one ⭐

PR #198, opened 2024-09-14
Added setting to disable automatic HotReload

CadeEvs:
- Moved to using an enum for the automatic reload developer setting for multiple states (`OnScriptSave`, `OnEditorFocus`, `Off`).
- Moved to using an enum for `bIsReloading` in `UnrealSharpEditor` for multiple states (`Inactive`, `PendingReload`, `Active`).
  - `PendingReload` is a new state that gets set to true when the DirectoryWatcher finds script changes, but the automatic setting is set to only Hot Reload on Focus or Manual.
- `HotReloadStatus` is now set to active and inactive within `StartHotReload()`, so it does not need to be manually set to true and false when calling the function.
  - `StartHotReload()` is no longer static so the `HotReloadStatus` can be set within the function.
- There's now static `OnCompileManagedCode()` and `OnOpenSolution()` functions that the commands register to, to match the other `On` prefixed static command functions.


OlssonDev:
Awesome feature ⭐ Thank you for the PR :D

PR #203, opened 2024-09-18
Recursive dependency copying

FailMelon:
Right now if there are native runtimes or any other files in sub directories in project bin folders they are ignored, this fixes that by copying files over recursively.


OlssonDev:
Thank you for the fix! 😄 ⭐

PR #204, opened 2024-09-18
Changed Delegates glue exposure

FailMelon:
Any delegates in a project the weaver will attempt to glue even if the user doesn't want to expose a delegate to USharp, this changes that and requires all delegates that the user wants to be glued requires an attribute.

```c#
// Multicast delegates.
[UMultiDelegate]
public delegate void MyShowcaseMulticastDelegate(int a);

// Single delegates.
[USingleDelegate]
public delegate void MyShowcaseDelegate(int a);
```

### DIFF
--- a/Managed/UnrealSharpPrograms/UnrealSharpWeaver/Program.cs
+++ b/Managed/UnrealSharpPrograms/UnrealSharpWeaver/Program.cs
@@ -255,8 +255,8 @@ public static class Program
             UnrealEnumProcessor.ProcessEnums(enums, metadata);
             UnrealInterfaceProcessor.ProcessInterfaces(interfaces, metadata);
             UnrealStructProcessor.ProcessStructs(structs, metadata, userAssembly);
-            UnrealDelegateProcessor.ProcessMulticastDelegates(multicastDelegates);
-            UnrealDelegateProcessor.ProcessSingleDelegates(delegates);
+            UnrealDelegateProcessor.ProcessMulticastDelegates(multicastDelegates, userAssembly);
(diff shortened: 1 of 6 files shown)


Managed/UnrealSharpPrograms/UnrealSharpWeaver/WeaverHelper.cs
FailMelon, line 1: Didn't realize that the weaver searched for the extension and not the type when doing delegate lookups, changed it so it'll now find delegate extension from the type name. If this is not wanted let me know and i'll revert the changes back to using the extension type lookup.

Managed/UnrealSharpPrograms/UnrealSharpWeaver/Program.cs
FailMelon, line 1: Didn't realize that the weaver searched for the extension and not the type when doing delegate lookups, changed it so it'll now find delegate extension from the type name. If this is not wanted let me know and i'll revert the changes back to using the extension type lookup.

OlssonDev:
Great PR! :D

Noticed a bug where the engine freeze if I add declare a TMulticastDelegate<> with a delegate that has no attribute.

FailMelon:
> Great PR! :D
> 
> Noticed a bug where the engine freeze if I add declare a TMulticastDelegate<> with a delegate that has no attribute.

Okay i pushed a revert to the extension changes

Managed/UnrealSharpPrograms/UnrealSharpWeaver/Program.cs
FailMelon: Reverted changes

PR #229, opened 2024-10-06
Fixed UStruct may mismatch cause Guid changed

ChenRenault:
Fixed UStruct may mismatch cause Guid
In Unreal Engine 5.4.4


OlssonDev:
Thank you for the PR ⭐

PR #230, opened 2024-10-06
Rename CSharpForUE module to UnrealSharpCore. Re name plugin to UnrealSharp

Duroxxigar:
Went with UnrealSharpCore because there is already a lot of references to "UnrealSharp" with strings. So this hopefully can alleviate some potential confusion. UnrealSharp should be the overall plugin, which has a bunch of C++ modules. I didn't touch the C# side, but there could be arguments made on clearing up that stuff as well. But it isn't as much of a priority as the idea is that people shouldn't need to mess with that side of things.


OlssonDev:
Thank you for the PR ⭐ Really needed!

PR #233, opened 2024-10-11
Multi Execute output pin support for BlueprintCallable Functions

JothamR:
Allow multiple execute flow via the ExpandEnumAsExecs="Parameter" for BlueprintCallable functions

Example use:
```
[UEnum]
public enum EnumResult { A, B, C }
```
```
[UFunction(FunctionFlags.BlueprintCallable, ExpandEnumAsExecs = "result")]
void MultiFlowTest(string input, out EnumResult result) 
```


OlssonDev:
Maybe this can be a UMetaData child attribute instead? I'm thinking so we don't clog UProperty

JothamR:
> Maybe this can be a UMetaData child attribute instead? I'm thinking so we don't clog UProperty

Sure, will take a look

JothamR:
oh amazing! Didn't realise all the additional fields were already supported via 
```
    [UMetaData("ExpandEnumAsExecs", "outcome")]
    [UMetaData("HideSelfPin","true")]
    etc.
```

So it's really just down to which ones look better on the [UFunction] attribute

OlssonDev:
I was thinking a UMetaData attribute class which is called ExpandEnumAsExecs, so you can just add it and get the meta data :D

JothamR:
Take 2 is now up, applied as individual attributes so you can tag using:
[HideSelfPin], [ExpandEnumAsExecs("outcome")], etc.

Full list of tags is here: https://dev.epicgames.com/documentation/en-us/unreal-engine/metadata-specifiers-in-unreal-engine

(Added a namespace `Unreal.Attributes.MetaTags` so it could find all the attributes tags together)

OlssonDev:
Awesome PR ⭐ Thank you!

PR #235, opened 2024-10-12
Fixed issue with manually exported types

FailMelon:
This PR fixes an issue currently with manually exported types not having the GeneratedType attribute applied to them causing them to become FallBack structs when doing object lookups by name C++ side.


OlssonDev:
Thank you for the PR⭐

PR #243, opened 2024-10-23
-fix usage of Quaternion instead of FQuat

rob-ack:
this fixes an issue in UnrealSharp.csproj FVector2D.Transform(FVector2D value, FQuat rotation) which had used Quaternion instead of FQuat in arg rotation


OlssonDev:
Thank you for the PR ⭐ !

PR #255, opened 2024-10-31
Add functionality to reload modules only & Add shortcuts

markusstephanides:
I added a quick functionality that let's users build the modules in their IDEs and then let UnrealSharp only weave and reload the modules instead of building them. This allows me to handle any errors of my code inside my IDE and let UnrealSharp only do the reload work when it's actually needed.

I also added two shortcuts so compiling & reloading can be accessed quickly.

If that change is ok I would also go ahead and implement automatic reloading if the module files changed to complete this additional workflow.

### DIFF
--- a/Source/UnrealSharpCore/CSDeveloperSettings.h
+++ b/Source/UnrealSharpCore/CSDeveloperSettings.h
@@ -9,6 +9,8 @@ enum EAutomaticHotReloadMethod : uint8
 {
 	// Automatically Hot Reloads when script changes are saved
 	OnScriptSave,
+	// Automatically Hot Reloads when the built .NET modules changed (build the C# project in your IDE and UnrealSharp will automatically reload)
+	OnModuleChange,
 	// Wait for the Editor to gain focus before Hot Reloading
 	OnEditorFocus,
 	// Will not Hot Reload automatically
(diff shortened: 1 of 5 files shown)


OlssonDev:
Finally could check it out! :D 

Nice feature! I'd love to see the shortcuts working.

markusstephanides:
Thanks, will look into why they are not working!

markusstephanides:
Shortcuts are fixed & added automatic reloading when modules were changed

CadeEvs:
Testing it right now and it looks like when AutomaticHotReloading is set to "OnModuleChange" it will hot reload twice. Once on module change and also when the editor gains focus

CadeEvs:
Just took a look. It looks like it triggers StartHotReload twice. Both stemming from OnCSharpCodeModified and due to the same managed dll. I also only have one project

markusstephanides:
Does it trigger twice immediately or with some time in between or directly after each other?

markusstephanides:
Stupid me, forgot to set the param for `StartHotReload` so it only reloads and not rebuild. Can you check again and let me know if it works now?

CadeEvs:
Works now! Great work

PR #256, opened 2024-11-01
Removed now unnecessary array fix

FailMelon:
Removed some changes from #147 that are no longer neccessary and gets rid of a warning that Unreal spams in console when using RPCs with arrays.


OlssonDev:
Thank you ⭐

PR #258, opened 2024-11-09
Code Generator fixes

rob-ack:
-Fix AsyncWrapperGenerator and NativeCallbacksWrapperGenerator do not pick up generic type argument namespaces which will lead to failing compilation of generated code

-Fix delegate parameter could not be passed by ref or in or out when used with TDelegate.
image


OlssonDev:
Awesome! Thank you for the PR ⭐

Issue #250, opened 2024-10-26
[BUG] Methods created in U# and called from non-parent blueprints using structs will cause blueprint errors


John-Storm:
Repo:
Create these classes and the struct in U#

```
[UClass]
public partial class UMyGameInstance : UGameInstance
{
    [UFunction(FunctionFlags.BlueprintCallable)]
    public FMyStruct GetData(FMyStruct updatedData)
    {
        updatedData.Counter++;
        updatedData.Name = "NewName";
        return updatedData;
    }
}
```
```
[UClass]
public class AMyActor : AActor
{
}
```
```
[UStruct]
public struct FMyStruct
{
    [UProperty(PropertyFlags.BlueprintReadWrite)]
    public int Counter;
    [UProperty(PropertyFlags.BlueprintReadWrite)]
    public string Name;
}
```

Create a blueprint called BP_MyActor that inherits from MyActor
Create a blueprint called BP_MyGameInstance that inherits from MyGameInstance
Fill the blueprints with content seen in the attached images
Compile the C# solution and observe the error highlighted in the second image

![image](https://github.com/user-attachments/assets/24e8e626-03d5-4700-af7b-578c41757d6b)
![image](https://github.com/user-attachments/assets/e24e6505-5885-4edd-83d0-ced15e31b29b)

LogBlueprint: Error: [AssetLog] C:\UnrealTestProjects\UnrealSharpTest\Content\BP_MyActor.uasset: [Compiler] Can't connect pins  My Struct  and  Return Value : Only exactly matching structures are considered compatible.
LogBlueprint: Error: [AssetLog] C:\UnrealTestProjects\UnrealSharpTest\Content\BP_MyActor.uasset: [Compiler] Can't connect pins  My Struct  and  Updated Data : Only exactly matching structures are considered compatible.
LogBlueprint: Error: [AssetLog] C:\UnrealTestProjects\UnrealSharpTest\Content\BP_MyActor.uasset: [Compiler] Can't connect pins  Updated Data  and  My Struct : Only exactly matching structures are considered compatible.
LogBlueprint: Error: [AssetLog] C:\UnrealTestProjects\UnrealSharpTest\Content\BP_MyActor.uasset: [Compiler] Can't connect pins  Return Value  and  My Struct : Only exactly matching structures are considered compatible.
[...]

PR #262, opened 2024-11-11
-fix #238 #250 reinstancing is not performed on method (UK2Node_CallFunction or UK2Node_CSAsyncAction) nodes.

### LINKED ISSUE #238: [BUG] Recompile Error with async function with return value
UE 5.4.4
UnrealSharp 0.1

1. BpExampleActor extends from AExampleActor. When AExampleActor is modified again, CS compilation generates an error.

image
LogBlueprint: Error: [AssetLog] F:\UEPrjs\TestUESharp\Content\Blueprints\BpExampleActor.uasset: [Compiler] This blueprint (self) is not a ExampleActor_C_OLD_119206, therefore ' Target ' must have a connection.

2. When clicking on UE's Compile button, it works fine.

image

rob-ack:
-properly reconstruct methods and async methods on c# recompile reload. this will also now pick up any changes to signatures and prevent blueprints from breaking since they reference their old types.

image

image

will fix #238 and #250

### DIFF
--- a/Source/UnrealSharpBlueprint/K2Node_CSAsyncAction.cpp
+++ b/Source/UnrealSharpBlueprint/K2Node_CSAsyncAction.cpp
@@ -25,6 +25,20 @@ UK2Node_CSAsyncAction::UK2Node_CSAsyncAction()
 	ProxyActivateFunctionName = GET_FUNCTION_NAME_CHECKED(UCSBlueprintAsyncActionBase, Activate);
 }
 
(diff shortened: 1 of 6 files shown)


OlssonDev:
General comments to follow UE's guidelines:

const Type - not Type const
Don't use auto
PascalCase

Otherwise it looks good :D

OlssonDev:
Thank you for the PR! Great fix ⭐

PR #268, opened 2024-11-13
[Minor] MetaTags - Suppress 'Parameter is unread' warnings on this file

JothamR:
Suppress 'Parameter is unread' on this file as these parameters are used via reflection
(and add ToolTip as a valid for Properties)


OlssonDev:
Thank you for the PR! :D ⭐

OlssonDev:
Awesome! I need to add the normal way of doing tooltips too! Thank you for the PR ⭐

PR #269, opened 2024-11-13
fix #250: BP skeleton updates where not performed on BP's using u# types (i.e. as function parameters or variables)

### LINKED ISSUE #250: [BUG] Methods created in U# and called from non-parent blueprints using structs will cause blueprint errors

rob-ack:
-replace more remaining data in BPS to update for reinstancing. also unreals api is used to trigger skeleton updates and changed events so recompilation will trigger like it would with c++  class changes.

FunctionOnBpTest
This is solving the case shown above. Using a BP class not inheriting from c# with a function using a struct from c# in its parameters and the BP will use a variable of type of the c# struct too,

### DIFF
--- a/Source/UnrealSharpEditor/Reinstancing/CSReinstancer.cpp
+++ b/Source/UnrealSharpEditor/Reinstancing/CSReinstancer.cpp
@@ -1,6 +1,7 @@
 ﻿#include "CSReinstancer.h"
 #include "BlueprintActionDatabase.h"
 #include "K2Node_CallFunction.h"
+#include "K2Node_StructOperation.h"
 #include "AssetRegistry/AssetRegistryModule.h"
 #include "UnrealSharpCore/TypeGenerator/Register/CSTypeRegistry.h"
 #include "Kismet2/BlueprintEditorUtils.h"
@@ -225,13 +226,61 @@ bool FCSReinstancer::UpdateMemberCall(UK2Node_CSAsyncAction * Node) const
 	return false;
 }
 
+void FCSReinstancer::UpdateInheritance(UBlueprint* Blueprint, bool& RefNeedsNodeReconstruction) const
+{
+	//we do not loop through the parent classes, as every blueprint will get an update on its own.
+	if(UClass* ParentClass = Blueprint->ParentClass)
+	{
+		if (UClass* const* NewClass = ClassesToReinstance.Find(ParentClass))
+		{
+			Blueprint->ParentClass = *NewClass;
+			RefNeedsNodeReconstruction = true;
+		}
+	}
+
+	for (FBPInterfaceDescription& InterfaceDescription : Blueprint->ImplementedInterfaces)
+	{
+		if (UClass* const* NewInterface = InterfacesToReinstance.Find(InterfaceDescription.Interface))
+		{
+			InterfaceDescription.Interface = *NewInterface;
+			RefNeedsNodeReconstruction = true;
+		}
+	}
+}
+
+void FCSReinstancer::UpdateNodePinTypes(UEdGraphNode* Node, bool& RefNeedsNodeReconstruction) const
+{
+	if (UK2Node_EditablePinBase* EditableNode = Cast<UK2Node_EditablePinBase>(Node))
+	{
+		for (const TSharedPtr<FUserPinInfo>& Pin : EditableNode->UserDefinedPins)
+		{
+			if (TryUpdatePin(Pin->PinType))
+			{
+				RefNeedsNodeReconstruction = true;
+			}
+		}
+	}
+
+	for (UEdGraphPin* Pin : Node->Pins)
+	{
+		if (TryUpdatePin(Pin->PinType))
+		{
+			RefNeedsNodeReconstruction = true;
+		}
+	}
+}
+
 void FCSReinstancer::UpdateBlueprints()
 {
-	TArray<UK2Node*> NodesToUpdate;
+	TSet<UBlueprint*> ToUpdate;
 	
 	for (TObjectIterator<UBlueprint> BlueprintIt; BlueprintIt; ++BlueprintIt)
 	{
+		bool bNeedsNodeReconstruction = false;
 		UBlueprint* Blueprint = *BlueprintIt;
+
+		UpdateInheritance(Blueprint, bNeedsNodeReconstruction);
+		
 		for (FBPVariableDescription& NewVariable : Blueprint->NewVariables)
 		{
 			TryUpdatePin(NewVariable.VarType);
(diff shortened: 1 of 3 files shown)


rob-ack:
You are mistaking. The function is implemented in the same BP not in the Parent as stated. 

According to your report i tested also the case you stated like shown here https://github.com/UnrealSharp/UnrealSharp/issues/250#issuecomment-2474940568:

John-Storm:
Confusing stuff! Looks like you've replicated the exact scenario. I do find some odd behavior if I use make struct, the error highlighting on the nodes goes away, and this remains as long as the project is open, even if I delete the make struct and goes back to the variable struct. The output log still reports the error regardless...

https://github.com/user-attachments/assets/ff5143fe-2a0f-4fe2-baaf-c6014ab71bbf

rob-ack:
The only thing i am thinking of is that you did not yet load the updated plugin code into your project.
Even though i think you know by yourself ensure the following:
- Git load up to commit b03985234f8539f272db6f60714cf00b7c74a3ce (https://github.com/rob-ack/UnrealSharp/commit/b03985234f8539f272db6f60714cf00b7c74a3ce)
- go to your project and run UBT (unreal build tool) like i stated in comment here https://github.com/UnrealSharp/UnrealSharp/issues/250#issuecomment-2474940568
this will update the plugin code. make sure the code actually compiled. 

try out the project.
if you still not seeing an update:
- make sure UBT actually build. its a bit tricky that it may silently fail. the user experience is not that great here.
- make sure you have no local plugin git repo modifications
- i have recently seen issues with Visual Studio V17.12.0. i had to revert back to 17.11.6. my unreal project was unable to build.
- you can try to load the VS Solution and run the editor from there. by this you have the best user feedback if the updated code is actually running and the project build.

John-Storm:
I created another project at my home computer, pulled UnrealSharp and set git to commit https://github.com/UnrealSharp/UnrealSharp/commit/b03985234f8539f272db6f60714cf00b7c74a3ce
(This is another test project I have at home, so the method names, BP names etc. doesn't match the other)

Everything seems to be compiled correctly and the source code looks to be the correct one. I'm able to attach the debugger and the breakpoint is hit for refreshing the blueprints after compiling.

https://github.com/user-attachments/assets/64f83c73-717d-449c-a218-585d38a6f80b

I'm sure it's something I completely misunderstand or did wrong... right now I'm at a loss. I'm throwing out some info just in case that might reveal something...

UE Ver:
image
System:
image
image

rob-ack:
Thank you. Oh boy so far my project does not reproduce the issue. 
But i recognized that a cast is involved and we have not now reinstating code for casts.
It must be the source of the errors.

Im adding it now.

I also want to share what we are actually looking for about this error for all to understand what and how so far:
Its about references to our old class types where for now after a recompile all our c# classes are marked as '_Old'.
It looks like this:
image

The hard stuff is to find all the nodes, which are of different types ofc, and replace such data.
As soon as such data is found i need to look into a way to replace it.

This is the way it was working until now but with many missing code for references. 
Maybe in the future we could do it completely differently but for now its done this way.

Ill push updates fixing the casts.

John-Storm:
The crazy thing is that the error doesn't always reproduce. If for some reason the error goes away, I must close the UE project and open fresh to have the error come back.

I appreciate the insight, and gave me an idea to try. Instead of casting directly, I used an interface, and that approach seems to work.

image

rob-ack:
@John-Storm 
Thank you for the feedback and the testing. I just pushed a new commit which now handles dynamic casts.
Also i found inconsistencies in the deep hierarchies of BP types deriving directly or indirectly from u# types (like shown above before).
For this i also added a UE API calls which will update those. I also found a way to reproduce via another completely new UE project . 🤕  

Pls try with the new commit. 
There still might be more stuff which we need to update on reinstancing. If you have more BP code which might be broke let me know. Thx.

John-Storm:
The latest commit solved the issue. Thanks for all the hard work!

image

rob-ack:
Nice. Thank you for the committed support :)

OlssonDev:
Awesome work tracking this bug! Thank you so much ⭐

PR #276, opened 2024-11-17
Adding more re instancing code for known collection types (map, array or set)

rob-ack:
-additional missing reinstancing code to support hot reload using u# types
-fix map, array, and set types
-fix interface function overloads which work with func refs:
image


OlssonDev:
Great work! Thank you for the PR ⭐

PR #277, opened 2024-11-19
U# additional analyzers

Tesagi:
Added new analyzers:
1. UObject new keyword creation
2. DefaultComponent on not UActorComponent types
3. DefaultComponent Property without setter

image


OlssonDev:
Great PR! This is really needed ⭐ ⭐ Thank you!

PR #278, opened 2024-11-19
[BUG, DEADLOCK] fix build tool may deadlock on reading weaver output and error streams in sequence

rob-ack:
this is when the weaver will run into errors. this is then deadlocking unreal during c# compilation.
also give UE thread a little rest time while waiting for recompilation process output


OlssonDev:
Thank you for the PR ⭐ Great stuff :D

PR #296, opened 2024-12-07
Fix #295 Prevent Deadlock during reflection data load. Prevent IL Weaver to generate colliding duplication entries.

### LINKED ISSUE #295: [DEADLOCK] Inheriting a C# class from a U# UInterface will deadlock during load
Inheriting like this.
image

But mainly we are endlessly looping through the following code and loading will not continue:
image

rob-ack:
fix #295

### DIFF
--- a/Managed/UnrealSharpPrograms/UnrealSharpWeaver/MetaData/BaseMetaData.cs
+++ b/Managed/UnrealSharpPrograms/UnrealSharpWeaver/MetaData/BaseMetaData.cs
@@ -1,9 +1,11 @@
-﻿using System.Globalization;
+﻿using System.Diagnostics;
+using System.Globalization;
 using Mono.Cecil;
 using Mono.Cecil.Cil;
 
 namespace UnrealSharpWeaver.MetaData;
 
+[DebuggerDisplay("{GetType().Name}: {Name}")]
 public class BaseMetaData
 {
     public string Name { get; set; }
--- a/Managed/UnrealSharpPrograms/UnrealSharpWeaver/MetaData/ClassMetaData.cs
+++ b/Managed/UnrealSharpPrograms/UnrealSharpWeaver/MetaData/ClassMetaData.cs
@@ -8,8 +8,8 @@ public class ClassMetaData : TypeReferenceMetadata
 {
     public TypeReferenceMetadata ParentClass { get; set; }
     public List<PropertyMetaData>? Properties { get; set; }
-    public List<FunctionMetaData> Functions { get; set; }
-    public List<FunctionMetaData> VirtualFunctions { get; set; }
+    public HashSet<FunctionMetaData> Functions { get; set; }
+    public HashSet<FunctionMetaData> VirtualFunctions { get; set; }
     public List<string> Interfaces { get; set; }
     public string ConfigCategory { get; set; } 
     public ClassFlags ClassFlags { get; set; }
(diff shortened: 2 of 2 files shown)


rob-ack:
@OlssonDev pls have a look on this proposal.

OlssonDev:
When does this happen? Isn't it better to see where we add it multiple times?

OlssonDev:
That limitation is stated here:
https://www.unrealsharp.com/documentation/functions/interfaces

rob-ack:
Yes. There are 2 Problems and we may work here a little more on the issue. 

1. It turned out that there is an issue when you use the Interface in U# and derive a type directly (not in PB). 

In this case you can not call into Properties. The problem lies in none created IL weaving initialization code.
The following image shows an exception and crash due to this problem.
image

This is because that IL code is only generated when you are using the UFunction attribute and the method is on the first implemented inheritance chain type defining the method. Otherwise its considered virtual and the static init IL code is not generated fot the type and only 

This in turn if you use UFunction declaration in the Interface only and not for the method in the first defining type, your function is considered virtual.

2. To solve this (or work around) i used the UFunction attribute on the method in the first declaring type. But then the UE will deadlock loading my assembly (which should ofc not happen). 

It turned out that this is because IL weaving now generated doubled fields for the UFunction methods like shown below.
image

But because i now use UFunction on the defining method the method is considered both [normal Function and also VirtualFunction](https://github.com/rob-ack/UnrealSharp/commit/cd40106252a5d65eb8ad6ada7235eea457e223e7#diff-f15b5f1fed69eda89e2203de33dffde33a26b7c0c5f7aa9b49a00449d87ce638L12).
This leads to an infinite loop during property iteration during loading the type in UE.
image

And investigation showed that with this usage the method is considered to run into both branches of [BaseMetaData.PopulateFunctions()](https://github.com/UnrealSharp/UnrealSharp/commit/cd40106252a5d65eb8ad6ada7235eea457e223e7#diff-01857e61a5965c144a4b7437fc0f3bc364dded205594bb7a97767777a9516ddfR8).
And this creates a FunctionMetaData instance twice which it seems was never intended and created the doubled IL code shown above.

**This PR provides Solution for these problems.**
1. Prevent a second FunctionMetaData in `BaseMetaData.PopulateFunctions()` [here](https://github.com/UnrealSharp/UnrealSharp/commit/cd40106252a5d65eb8ad6ada7235eea457e223e7#diff-01857e61a5965c144a4b7437fc0f3bc364dded205594bb7a97767777a9516ddfR8)
and [here](https://github.com/UnrealSharp/UnrealSharp/commit/8655d8547ebba5842ecb98a0a82ee5188e64b2bc#diff-f15b5f1fed69eda89e2203de33dffde33a26b7c0c5f7aa9b49a00449d87ce638R126)
To never generate double IL code.

2. With change 1 applied i was able to just use UFunction on my declaring type method even though not intended possibly but working. This way no additional code change where required to make my code work and crash or deadlock the editor.

My conclusion is that having the function as both, normal Function and also VirtualFunction was never intended so this is accounted for.

**Additionally**
Changing the code so that UFunction is not required on the first defining method is possible but was not what this PR is about so far.
@OlssonDev But if we agree i can extend this PR to do this as well and add code to check for this like it seems it was actually intended to do.

OlssonDev:
https://github.com/UnrealSharp/UnrealSharp/commit/5359c7832553040b104feb22bcf25d2eb91c9a1a 

Fixed here. Please verify on your end :D

rob-ack:
Can confirm it works with 35e6f7f14e29cf28df020763b65ade7c3394ba3e ;)
GJ

Issue #249, opened 2024-10-26
[BUG] Opening the C# solution from the U# menu will freeze the UE Editor


John-Storm:
![image](https://github.com/user-attachments/assets/db9bbb1f-f17d-45a3-816d-9a3a9e6c99b2)

UE 5.4.4: When clicking that link, the C# solution will open but the UE Editor hangs. Opening the <MyManagedProject>.sln outside of using this shortcut doesn't cause any issue with the UE Editor.

Tested on a new blank game project.

Zuriel-Zurichi:
I don't know if this helps but this only happens to me when I use Visual Studio

John-Storm:
Interesting observation. I use Visual Studio Community 2022. Didn't try with other IDEs.

rob-ack:
I can not confirm this. It is not happening for me. 
I have VS 2022 community too with Version 17.11.5 currently.

ezesil:
Same here using Visual Studio Community 2022 17.10.5 and Unreal Engine 5.4.2 compiled from source. After closing it, it stops freezing. Opening visual studio from outside has no issues. 

It's probably blocking unreal's main thread.

NGC717:
Using UE 5.4.4 and Rider 2024.2.7, encountered the same issue.

349516385:
also happens to me

JothamR:
Looks like it blocks in UnrealSharpEditor.cpp OpenSolution() on the line:
`FPlatformProcess::ExecProcess(TEXT("cmd.exe"), *OpenSolutionArgs, nullptr, nullptr, nullptr);`

JothamR:
Doesn't look like there's a parameter to make ExecProcess non-blocking, but we could run it as a separate AsyncTask or thread.

As an AsyncTask it'd change to this:
```
	AsyncTask(ENamedThreads::AnyBackgroundThreadNormalTask, [OpenSolutionArgs]() 
	{
		FPlatformProcess::ExecProcess(TEXT("cmd.exe"), *OpenSolutionArgs, nullptr, nullptr, nullptr);
	});
```

JothamR:
I can create an PR with the above as it seems to fix the issue, but happy for someone who knows more about how multi-threading in Unreal works to weigh in.  (e.g. pros/cons of using AsyncTask vs a new thread (via FRunnable?) vs a more standard C++ thread)

JothamR:
[...]

PR #297, opened 2024-12-08
In Menu -> Open C# Solution, open using CreateProc instead of ExecProcess (as ExecProcess is a blocking call)

JothamR:


JothamR:
for this issue https://github.com/UnrealSharp/UnrealSharp/issues/249

OlssonDev:
Thank you for the fix ⭐ !

PR #304, opened 2024-12-12
Check for missing launchSettings.json in C# project files during UE startup and add it if absent.

meglinge:
When managing a project with Git, it is necessary to ignore the Properties/launchSettings.json file in all .NET projects within the Script folder because this file stores the absolute paths of Unreal Engine and the project on the local machine. However, the Properties/launchSettings.json file is only generated when a C# project is initially created. Therefore, a mechanism is needed to detect all .NET projects in the Script folder (excluding ProjectGlue) upon each project startup and automatically add the Properties/launchSettings.json file to the ignore list.


OlssonDev:
I'd like to see a fix where we can minimize duplicated code! Otherwise great fix! ⭐

meglinge:
> I'd like to see a fix where we can minimize duplicated code! Otherwise great fix! ⭐我希望看到一个修复，可以最小化重复代码！否则非常好的修复！ ⭐

fixed

PR #308, opened 2024-12-14
Defensive code to stop a crash on Reinstancing in Hot-Reload

JothamR:
Defensive code to stop a crash on Hot-Reload when an old object is marked for garbage collection while it is still `IsRooted`
(`Old->MarkAsGarbage()` does an assertion on this and crashes when the check fails)

Will write up an issue to match, since there's likely an underlying issue that caused the bad state, but this helps stop the editor from crashing, adds some logging on which object has the problem, and let's you continue.


JothamR:
For issue: https://github.com/UnrealSharp/UnrealSharp/issues/309

OlssonDev:
The objects shouldn't be rooted. So this is very strange

OlssonDev:
https://github.com/UnrealSharp/UnrealSharp/pull/311

I fixed crashes related to structs here

OlssonDev:
This has been fixed in another PR! So I close this, but thank you for the work ⭐

PR #313, opened 2024-12-15
Fixes WorldContext not being set during Async

FailMelon:
Async delegates do not have WorldContext set causing errors for anything that requires it, this PR fixes that.


OlssonDev:
Thank you for the PR ⭐ !

PR #314, opened 2024-12-16
Fixes certain property access modifiers

FailMelon:
This PR fixes an issue where certain properties were not correctly getting the highest getter and setter access modifier, causing the property to become protected when it should be public.


OlssonDev:
Thank you for the PR ⭐ !

PR #322, opened 2024-12-22
Fixes FName data alignment

FailMelon:
This issue is a quick fix for FName during packaged builds mainly for the new logging system.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/Extensions/CoreUObject/Object.cs
+++ b/Managed/UnrealSharp/UnrealSharp/Extensions/CoreUObject/Object.cs
@@ -10,7 +10,19 @@ public partial class UObject
     /// <summary>
     /// The name of the object in Unreal Engine.
     /// </summary>
-    public FName ObjectName => IsDestroyed ? FName.None : UObjectExporter.CallNativeGetName(NativeObject);
+    public FName ObjectName
+    {
+        get
+        {
+            if (IsDestroyed)
+            {
+                return FName.None;
+            }
+            
+            UObjectExporter.CallNativeGetName(NativeObject, out FName objectName);
+            return objectName;
+        }
+    }
     
     /// <summary>
     /// Whether the object is valid. UObjects can be valid but pending kill.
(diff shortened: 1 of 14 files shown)


Liangjia0411:
If using the Debug version at runtime, should there still be errors?
image

Liangjia0411:
The types of the three variables in FName should be uint instead of int

Liangjia0411:
When I add the DEBUG macro to the "DisplayIndex" field, getting FName during packaged builds will still crash;@FailMelon

When calling UUObject Exporter:: NativeGetName, the passed `Object` value is incorrect.

The value of `Object` is the same as the value of register `rdx`, but the actual value should be the value of register `rcx`

Wrong value from rdx
  image
  image


Correct value from rcx
image

FailMelon:
Yeah i'd rather not use DEBUG i think having a custom preprocessor called PACKAGE or something, was just a quick fix for anyone that needed it for the new logging system, i know the 3 variables are uints it doesnt matter. Not sure what issue your getting with the Object being passed in im not getting that error in my project, how are you using FName?

FailMelon:
Added a PACKAGE preprocessor directive instead of using DEBUG now

Liangjia0411:
I am simply obtaining the name of UObject
image
image

Liangjia0411:
> When I add the DEBUG macro to the "DisplayIndex" field, getting FName during packaged builds will still crash;@FailMelon
> 
> When calling UUObject Exporter:: NativeGetName, the passed `Object` value is incorrect.
> 
> The value of `Object` is the same as the value of register `rdx`, but the actual value should be the value of register `rcx`
> 
> Wrong value from rdx image image
> 
> Correct value from rcx image

Not only CoreCLR, I also encountered this question on Mono

FailMelon:
Ah yeah i also was having issues with this and ObjectName, it has to be looked into futherer as to why that happens this was mainly to address an immediate issue of data alignment with packaged products not working at all because of the new logger. You can use SystemLibrary to grab object names for now as a string, Anything you can find to help figure out why its happening for ObjectName would be appreciated i havn't had the time to look into it. I'm not sure at the moment if ObjectName is related to data alignment or not from what i've tested it seems to be coming from some other issue.

Liangjia0411:
This should be an API compatibility issue.C # FName and C++ FName are incompatible

PR #326, opened 2025-01-02
UInterface analyzer

Tesagi:
https://github.com/UnrealSharp/UnrealSharp/pull/325/commits/4133bf2ef5fb5e9674c8466f953ac4fc4263f3f4

Added 2 analyzer for UInterface attribute: 
1. UProperty with interface type
2. UFunction with interface parameter type

image

image

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp.SourceGenerators/AnalyzerStatics.cs
+++ b/Managed/UnrealSharp/UnrealSharp.SourceGenerators/AnalyzerStatics.cs
@@ -1,4 +1,5 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


OlssonDev:
Awesome :D Thank you for the PR!

I see an edge case with this code, if you're using IList, IDictionary, et.c. which we support for containers. You'll get an error with this setup!

Tesagi:
I forgot about that, then need a list of interfaces and I'll add them to ignored in analyzer

OlssonDev:
Here they are:

```
IList
IReadOnlyList
IDictionary
IReadOnlyDictionary
ISet
IReadOnlySet
```

Tesagi:
Since IReadonlySet does not exist in the .net standard, I added a check by namespace + interface name

OlssonDev:
Thank you for the PR, love me some analyzers! ⭐

PR #331, opened 2025-01-17
Fixes world context issues

FailMelon:
This issue fixes some edge case issues with world context for now by removing the nullptr of the current context.


OlssonDev:
Verified! ⭐ Thank you for the fix and discussion :D

PR #334, opened 2025-01-24
Fixes autocast from causing some methods to be ignored.

FailMelon:
This pr fixes autocast from exporting functions if they don't pass the autocast checks.


OlssonDev:
Thank you for the fix ⭐

PR #338, opened 2025-01-24
Implicit boolean operator for UObjects

FailMelon:
This pr allows for implicit boolean operations on UObjects so you can check if they valid as a bool in an if statement similar to how you can in Unreal.


OlssonDev:
Thank you for the PR ⭐

PR #339, opened 2025-01-24
Auto generates StructLayout on blittable types

FailMelon:
This pr auto generates StructLayout on blittable type no longer requiring adding it manually, this also fixes potential issue with compiler optimizing the data layout of a blittable type in a way that would cause problems.


OlssonDev:
Thank you for the PR ⭐

PR #388, opened 2025-04-05
Fixes an issue with default tick settings not reapplying after hotreload

### LINKED ISSUE #383: Inherited PlayerController disables ticking once editor is reloaded
Looks like I have found a bug. I created a simple base class for PlayerController on the C# level. 
``` csharp
[UClass]
public class APlayerControllerCore : APlayerController
{
   
}
```
Created a BLPlayerController on the blueprint level based on the APlayerControllerCore. Set this player controller to game mode. Now if I start the game all my inputs mouse and keyboard are ignored. 

It happens because BLPlayerController sets the "Start with Tick Enabled" to false. 
Image

You can set it to true manually and now all inputs are working fine. 

Also, you can press "Force Compile C#", open BLPlayerController and just press "Compile"(yes it is green, just press it). Now editor show "Start with Tick Enabled" as if it was modified:

Image

If you reset it(effectively changing to true) input will start working as in first case. Until you close the editor and reopen it. After that the "Start with Tick Enabled" goes back to false.

I have tried to modify APlayerControllerCore like that but it does nothing:
``` csharp
[UClass]
public class APlayerControllerCore : APlayerController
{
    public APlayerControllerCore()
    {
        ActorTickEnabled = true;
    }

    protected override void BeginPlay()
    {
        ActorTickEnabled = true;
        base.BeginPlay();
    }

    public override void ConstructionScript()
    {
        ActorTickEnabled = true;
        base.ConstructionScript();
    }
}
```

FailMelon:
This issue fixes default tick settings not being reapplied to the new CDO after a hot reload.

Fixes #383

### DIFF
--- a/Source/UnrealSharpCore/TypeGenerator/Register/CSGeneratedClassBuilder.cpp
+++ b/Source/UnrealSharpCore/TypeGenerator/Register/CSGeneratedClassBuilder.cpp
@@ -85,7 +85,8 @@ void FCSGeneratedClassBuilder::UpdateClassDefaultObject() const
 	
 	OwningAssembly->RemoveManagedObject(ClassDefaultObject);
 	
-	Field->GetDefaultObject(true);
+	UObject* DefaultObject = Field->GetDefaultObject(true);
+	SetupDefaultTickSettings(DefaultObject, Field);
 }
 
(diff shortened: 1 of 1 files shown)


OlssonDev:
Thank you for the fix :D

PR #389, opened 2025-04-08
Fixed IndexOf() crash with MapReadOnly

CadeEvs:


OlssonDev:
Thank you for the fix :D

PR #393, opened 2025-04-12
Added optional Key parameter to PrintString

CadeEvs:
Adds the key parameter to PrintString to have feature parity with the BP/CPP version


OlssonDev:
Thank you for the PR :D

PR #394, opened 2025-04-14
Fixed SourceGenerator for ActorComponents

CadeEvs:
The function must've changed to require a generic, so I've updated the source generator to fix a compile error when trying to use it with actor components


OlssonDev:
Thank you for the PR :DD

PR #395, opened 2025-04-15
Added validation checks to metadata fields.

FailMelon:
This PR fixes some json errors when parsing UnrealSharp.metadata.json. Since the metadata is a different kinda of data structure it fails to get the normal fields in a user assembly.


FailMelon:
#397 Closing this for the preferred fix

PR #396, opened 2025-04-15
Add new AddComponentByClass function

CadeEvs:
This function uses a generic for the return value and takes in a TSubclassOf. This is in cases where you have a base class you want to return as, but want to create a specific subclass using a TSubclassOf


OlssonDev:
Thank you for the PR! :D

PR #397, opened 2025-04-15
Fixes issues with loading UnrealSharp.metadata.json

FailMelon:
This PR fixes some json errors when parsing UnrealSharp.metadata.json. Since the metadata is a different kinda of data structure it fails to get the normal fields in a user assembly.


OlssonDev:
Thank you for the PR :D

PR #400, opened 2025-04-17
Adds support for marshalling structs using `CppStructOps::Copy`, exports `FMoverDataCollection` using this functionality

Y0SH1M4S73R:
`FMoverDataCollection` is a non-POD data type due to it having virtual methods (at least `StaticStruct`). However, it can be copied using `CppStructOps::Copy`, and has to be, due to possessing a non-reflected member property that is relevant for reflection-exposed functions. This PR adds support for marshalling structs by performing a native cpp struct ops copy, and adds that support to `FMoverDataCollection`.


OlssonDev:
Really cool feature! Thank you for the PR :D

PR #404, opened 2025-04-20
Adds OnHotReload callback support

EvieMae:
Enables UClasses to implement OnHotReload method, which is called after hot reload.

This allows for code to be executed after code changes are applied during runtime.

I've tested it on Pawns, Actors and UObjects, this is called per instance of object during RUNTIME only.  

```csharp
    [UFunction]
    public void OnHotReload()
    {
        Logger.Log($"UOBJECT {GetType().Name} Hot Reloaded");
    }
```

### DIFF
--- a/Source/UnrealSharpCore/CSAssembly.cpp
+++ b/Source/UnrealSharpCore/CSAssembly.cpp
@@ -368,6 +368,21 @@ UClass* FCSAssembly::FindClass(const FCSFieldName& FieldName) const
 	return Class;
 }
 
+
+TArray<UClass*> FCSAssembly::GetAllClasses() const
+{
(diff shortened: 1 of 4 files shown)


FailMelon:
Was wondering if we could approach this in a different way? Looping through all classes in a project and then all instances seems like it might be a bit of a performance hit for a uncommon feature. Perhaps some kind of listenable event would work better?

OlssonDev:
I agree with FailMelon. This approach would degrade the hot reload performance, a delegate would fit much better!

OlssonDev:
> For sure but implementing said method wouldn't subscribe right as it would need to be on initialize, unless that's the limitation

I'm not really following. Can't you subscribe on BeginPlay or the ctor?

PR #410, opened 2025-04-23
Natively-called c# Overrides Properly Copy Output Params

Y0SH1M4S73R:
`CSFunctionFactory::CreateOverriddenFunction` did not apply the `HasOutParms` function flag to overrides of functions that do.
This prevented `UObject::ProcessEvent` from creating an output property chain when invoking such functions. This is relevant for c# implementations of UInterface functions, as `ProcessEvent` is part of the code path used to execute non-native implementations of said functions.

Additionally, `CSFunction_Params::InvokeManagedMethod_Params` uses the output property chain of the passed in function (if present) when the function is being called from native code. This allows native invocations to correctly receive the output params of the c# function.


OlssonDev:
Thank you for the PR :D Awesome fix!

PR #416, opened 2025-05-11
Input Actions can be unbound from c#

Y0SH1M4S73R:
This PR allows input actions to be unbound after binding them. To do this, `UEnhancedInputComponent.BindAction` now has an `out uint` parameter that returns the binding's unique handle. The old signature is now an overload that discards the handle parameter. You can pass the output handle to `UEnhancedInputComponent.UnbindAction` to unbind the Action from the Input Action. Both `BindAction` and `UnbindAction` now return `bool`, specifying whether their respective operations were successful.


OlssonDev:
Awesome :D Thank you for the PR

PR #419, opened 2025-05-14
1 fix compile error on mac

meilaopo:
1 fix compile error on mac
2 fix marshalling on mac


OlssonDev:
Thank you for the fix :D

PR #423, opened 2025-05-21
Fix for: Failed to load UnrealSharpCompiler.dll

rob-ack:
-remove outdated .sln file not in use
-fix Failed to load UnrealSharpCompiler.dll. Unreal Sharp currently depends on StateTree since UnrealSharpCompiler.dll depends on it. So link StateTree.
```
3>EXEC : warning : Plugin 'UnrealSharp' does not list plugin 'StateTree' as a dependency, but module 'UnrealSharpCompiler' depends on module 'StateTreeModule'.

[2025.05.21-15.19.04:332][  0]LogWindows: Failed to load 'H:/ProjectsUnreal/XXX/Plugins/UnrealSharp/Binaries/Win64/UnrealEditor-UnrealSharpCompiler.dll' (GetLastError=126)
[2025.05.21-15.19.04:332][  0]LogWindows:   Missing import: UnrealEditor-StateTreeModule.dll
```


OlssonDev:
Thank you for the fix and cleanup :D

PR #426, opened 2025-05-22
Added memcpy fallback for NativeCopy

FailMelon:
This PR adds a fallback memcpy if the CppStructOps doesn't have the capability of copying.


OlssonDev:
Thank you for the PR :D

PR #427, opened 2025-05-23
-fix native and managed function size mismatch due to wrong parameter definition

rob-ack:
fix for the NativeFunction.Size (48) and ManagedFunctionSize (32) size mismatch.


OlssonDev:
Thank you for the fix! :D

PR #429, opened 2025-05-25
Update AsyncWrapperGenerator.cs

blizzzrdda:


OlssonDev:
Ah great catch, thank you for the fix! :D

PR #430, opened 2025-05-27
Fixes issue with opening vs solution in U#

FailMelon:
This PR fixes an issue with opening solution in U# with the wrong msbuild targets, haven't tested to see if this still works fine with rider.


OlssonDev:
Thank you for the fix :D

PR #437, opened 2025-06-01
Fixed LoadPrimaryAssetClass

blizzzrdda:
`if (loaded is not T typedObject)
{
    throw new InvalidCastException($"Asset '{assetId}' could not be cast to '{typeof(T).Name}' or was not found.");
}
`

this caused editor crash when use AssetManager to load Blueprint. 

Instead, `SystemLibrary.GetClass` should be used to load the class, and compare to `typeof<T>`


OlssonDev:
Great catch! I've been at UnrealFest, sorry for the delay! Thank you for the fix

PR #447, opened 2025-06-17
fix compile error on macOS

RPG3D:


OlssonDev:
Thank you for the fix! :D

PR #449, opened 2025-06-20
Fix for "Unable to clear TMap" & Fix for TSet trying to bind none existing function "GetScriptSetLayout"

GameMill:


OlssonDev:
Thank you for the fix! :D

PR #450, opened 2025-06-20
unreal engine define constant inheritance on build

Tesagi:
Added unreal engine define constant inheritance on build.


OlssonDev:
Sorry for the slow response! Busy schedule lately!

Awesome PR, didn't know you could do these tricks, TIL.

PR #461, opened 2025-06-29
Fixed stack frame corruption when calling to C# from Blueprints

retroandchill:
Fixed an issue where the location of the local variables in the stack frame is changed and not restored before the call ends, leading to stack corruption when making additional calls.


OlssonDev:
Thank you for the fix! :D

PR #462, opened 2025-06-29
Fix alphabetical struct processing issue where a struct could be processed before its dependency.

GameMill:
Fix for following error
"
Error during assembly processing: Could not find marshaller class FPointMarshaller for type ManagedTesting.FPoint
System.AggregateException: Assembly processing failed (Could not find marshaller class FPointMarshaller for type ManagedTesting.FPoint)
 ---> System.Exception: Could not find marshaller class FPointMarshaller for type ManagedTesting.FPoint
".

 
The issue occurs because FContainerData is processed before FPoint, preventing the processor from locating the corresponding Marshaller class.

Test code below
ContainerData.cs
```
using UnrealSharp.Attributes;

namespace ManagedTesting
{
    [UStruct]
    public struct FContainerData
    {
        [UProperty(PropertyFlags.BlueprintReadOnly)]
        public IList<FPoint> X;
    }
}
```

Point.cs
```
using UnrealSharp.Attributes;

namespace ManagedTesting
{
    [UStruct]
    public struct FPoint
    {
        [UProperty(PropertyFlags.BlueprintReadOnly)]
        public int X;

        [UProperty(PropertyFlags.BlueprintReadOnly)]
        public int Y;

        [UProperty(PropertyFlags.BlueprintReadOnly)]
        public bool IsValid;
    }
}
    ```

### DIFF
--- a/Managed/UnrealSharpPrograms/UnrealSharpWeaver/TypeProcessors/UnrealStructProcessor.cs
+++ b/Managed/UnrealSharpPrograms/UnrealSharpWeaver/TypeProcessors/UnrealStructProcessor.cs
@@ -1,4 +1,4 @@
-﻿using Mono.Cecil;
+using Mono.Cecil;
 using Mono.Cecil.Cil;
 using UnrealSharpWeaver.MetaData;
 using UnrealSharpWeaver.Utilities;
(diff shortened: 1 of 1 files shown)


OlssonDev:
Thank you for the fix <3

PR #463, opened 2025-06-30
Added support for exporting NetMode to U#

GameMill:


OlssonDev:
ENetMode is already a reflected enum, so no need to create it again in C#. You can expose the getter through UFUNCTION

GameMill:
image
There is not ENetMode in C#.

GameMill:
There is EPlayNetMode but this is missing DedicatedServer option because it`s from the old UnrealEd 
image

OlssonDev:
Oh, I read it wrong. 

The enum could be blittable, you don't need to cast and return numerals. Can just return the enum from c++ to c#

GameMill:
@OlssonDev I`ve update the code for directly exporting the enum instead of using a switch.

the only issue I can see with direct export is depending on the C++ compiler for the some platform the underlaying type for the enum might be u_int8 and not int32.

OlssonDev:
I don't think we need to worry about it, most modern compilers will default to an int32. But great catch, which we'll keep in mind.

OlssonDev:
Thank you for the contribution <3

PR #464, opened 2025-07-01
Added support for using custom getters/setters for properties

### LINKED ISSUE #135: Support properties with custom getters and setters
The getter and setter implementation would be moved to ufunctions and linked to the unreal uproperty via `BlueprintGetter` and `BlueprintSetter`.

For example, the following native C++ code could be reproduced in C# as so:
```C++
UPROPERTY(EditAnywhere, BlueprintGetter = IsPassEnabled, BlueprintSetter = SetPassEnabled )
bool bEnabled = true;

UFUNCTION(BlueprintGetter)
bool IsPassEnabled() const;

UFUNCTION(BlueprintSetter)
void SetPassEnabled(bool bSetEnabledTo = true);
```
```C#
private bool isEnabled;

[UProperty(PropertyFlags.EditAnywhere)]
public bool Enabled
{
    get => isEnabled;
    set => isEnabled = value;
}
```

The weaver or sourcegen would turn the above property into something like this:
```C#
private bool isEnabled;

[UProperty(PropertyFlags.EditAnywhere, BlueprintGetter: nameof(Get_Enabled_Implementation), BlueprintSetter: nameof(Set_Enabled_Implementation)]
public bool Enabled
{
    get
    {
        // Native interop code here
    }
    set
    {
        // Native interop code here
    }
}

[UFunction(FunctionFlags.BlueprintGetter)]
private bool Get_Enabled_Implementation() => isEnabled;

[UFunction(FunctionFlags.BlueprintSetter)]
private void Set_Enabled_Implementation(bool value) => isEnabled = value;
```

Possibly something of note is that the original property would become inaccessible and unusable as a backing field - in the C++ example, native code can still access the original `bEnabled` field and use it as a backing field, whereas in C# you'd need to introduce another field (or uproperty for proper serialisation and GC tracking) as a backing field.

retroandchill:
Closes #135 

This has been an outstanding feature request that is going to help reduce a fair amount of boilerplate.
 
The following changes have been applied:

1. UProperties are now allowed to have custom getters/setters
2. A new attribute field has been added to specify the name of the backing field. If left empty, it will default to the compiler-generated backing field name, which should allow for the use of the [upcomming field keywork in C# 14](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/field-keyword)
3. If a property has either a custom getter or custom setter-defined the getter and setter are marked a UFunctions with BlueprintCallable and BlueprintInternalUseOnly, and those functions are automatically asigned to the getter and setter.
4. The backing field is replaced like it was before but now when replacements are made it scans the entire class for field access and replaces those with calls to accessing a new weaver generated property that contains the native interop. This way the user should be able to use the field within the class internals for access.

There only known limitation of this that I am aware of when debugging if you hover over the backing field it doesn't claims that it is unknown in the current context. I'm not sure if we need to make additional changes to the debug symbols to handle that gracefully, or if the generated property should share the name of the field.

### DIFF
--- a/Managed/UnrealSharpPrograms/UnrealSharpWeaver/TypeProcessors/PropertyProcessor.cs
+++ b/Managed/UnrealSharpPrograms/UnrealSharpWeaver/TypeProcessors/PropertyProcessor.cs
@@ -3,6 +3,7 @@ using Mono.Cecil.Cil;
 using Mono.Cecil.Rocks;
 using UnrealSharpWeaver.MetaData;
 using UnrealSharpWeaver.NativeTypes;
+using UnrealSharpWeaver.Utilities;
 
 namespace UnrealSharpWeaver.TypeProcessors;
 
@@ -40,11 +41,24 @@ public static class PropertyProcessor
             
             if (prop.MemberRef.Resolve() is PropertyDefinition propertyRef)
             {
-                prop.PropertyDataType.WriteGetter(type, propertyRef.GetMethod, loadBuffer, nativePropertyField);
-                prop.PropertyDataType.WriteSetter(type, propertyRef.SetMethod, loadBuffer, nativePropertyField);
+                if (prop.HasCustomAccessors)
+                {
+                    // For custom accessors, create a private property with marshalling code
+                    string backingFieldName = RemovePropertyBackingField(type, prop);
+                    PropertyDefinition privateProperty = CreatePrivateAccessorProperty(type, prop, propertyRef, loadBuffer, nativePropertyField);
+                    prop.GeneratedAccessorProperty = privateProperty;
 
-                string backingFieldName = RemovePropertyBackingField(type, prop);
-                removedBackingFields.Add(backingFieldName, (prop, propertyRef, offsetField, nativePropertyField));
+                    removedBackingFields.Add(backingFieldName, (prop, privateProperty, offsetField, nativePropertyField));
+                }
+                else
+                {
+                    // Standard property handling
+                    prop.PropertyDataType.WriteGetter(type, propertyRef.GetMethod, loadBuffer, nativePropertyField);
+                    prop.PropertyDataType.WriteSetter(type, propertyRef.SetMethod, loadBuffer, nativePropertyField);
+
+                    string backingFieldName = RemovePropertyBackingField(type, prop);
+                    removedBackingFields.Add(backingFieldName, (prop, propertyRef, offsetField, nativePropertyField));
+                }
             }
             
             prop.PropertyOffsetField = offsetField;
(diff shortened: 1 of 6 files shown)


retroandchill:
This PR also includes commits from #461, so that may need to be reviewed first. I branched from there instead of main because it's difficult to test things without the fix provided from that.

OlssonDev:
This is awesome! Although BlueprintGetter/Setter don't get called when the value is fetched through reflection, only from BP. The proposed solution for this is to use `TPropertyWithSetterAndGetter` which reflection can read from too.

retroandchill:
So would that mean it would fetch the memory that the generated property would return rather than the getter/setter? I also presume this is the same behavior that you see when you use BlueprintGetter/BlueprintSetter in C++?

OlssonDev:
Exactly, BlueprintSetter/Getter is purely for Blueprints. You'd need the `TPropertyWithSetterAndGetter ` to work with systems that fetch data through reflection

retroandchill:
That type seems pretty nested in the internals, I'm going to set a breakpoint and see what happens when a C# property is created to verify if it gets wrapped correctly or not

OlssonDev:
That type will call C# to marshal back to native was my plan.

retroandchill:
I can take a crack at that when I get home. I think we just need to do some static registration for all the property types to call a wrap method that is determined based on the type.

retroandchill:
Okay I just updated the setup to handle the creation of properties that will invoke the getter and setter during reflection access. I ended up creating a parallel structure for TPropertyWithSetterAndGetter that will invoke the UFunction for the getter and setter.

OlssonDev:
I can't seem to successfully weave cropout with the latest changes :/

retroandchill:
I'll try to pull it down and see if I can debug the issue with it

retroandchill:
Okay I just pushed up a fix

Managed/UnrealSharpPrograms/UnrealSharpWeaver/TypeProcessors/PropertyProcessor.cs
```diff
@@ -53,24 +67,25 @@ public static void ProcessClassMembers(
         RemoveBackingFieldReferences(type, removedBackingFields);
     }
     
-    static void RemoveBackingFieldReferences(TypeDefinition type, Dictionary<string, (PropertyMetaData, PropertyDefinition, FieldDefinition, FieldDefinition?)> strippedFields)
+    private static void RemoveBackingFieldReferences(TypeDefinition type, Dictionary<string, (PropertyMetaData, PropertyDefinition, FieldDefinition, FieldDefinition?)> strippedFields)
     {
-        foreach (MethodDefinition? constructor in type.GetConstructors().ToList())
+        foreach (MethodDefinition? method in type.GetConstructors().Append(type.GetMethods()))
```
OlssonDev, line 72: This will increase the hot reload time by a pretty large margin. Are there any better ways we can handle this?

retroandchill: If we allow the user to set the backing field directly I don't think so. It may be better to create an analyzer that forbids the setting of the backing field outside of the constructor and the property methods, requiring the user to always go though the property getter/setter, which may be for the best anyways, and if you need a property without a backing field you just need to mark the getter/setter as a [UFunction] instead (this is already possible).

OlssonDev: > If we allow the user to set the backing field directly I don't think so. It may be better to create an analyzer that forbids the setting of the backing field outside of the constructor and the property methods, requiring the user to always go though the property getter/setter, which may be for the best anyways, and if you need a property without a backing field you just need to mark the getter/setter as a [UFunction] instead (this is already possible).

The custom getter/setter should allow for backing fields. Native should get the value from managed memory whenever it tries to fetch the value

retroandchill: So the problem with that is that it works great when using reflection to get the value and write it to a pointer, but less so when altering the property in the editor, as many of those implementations rely on getting the physical address of the backing field in native memory, which will cause issues with non-blittable types.

retroandchill: Okay, I just pushed an update that will call to C# in all circumstances. It will do this by essentially syncing with the native memory for the property on calling the getter and setter using reflection as well as during serialization and deserialization and then letting the engine serialize the native memory.

PR #470, opened 2025-07-03
Allow for interfaces implemented in Blueprints to be called from C# code

### LINKED ISSUE #289: U# has no way to properly work with U# interfaces from within U# code called or implemented by blueprints.
1. TScriptInterface<> equivalent is missing: https://dev.epicgames.com/community/snippets/003/unreal-engine-interfaces-for-c-and-blueprint
2. Calls to Blueprint only implemented interface functions can not be performed: https://forums.unrealengine.com/t/c-interface-cast-returns-null-when-implemented-with-editor/334784/10

retroandchill:
Closes #289 

Fixes the problem of Blueprint-only interfaces by adjusting the way marshalling is handled to fall back when it's unable to properly cast to generate a wrapper for the interface. These wrappers implement the new IScriptStruct interface which is used when marshalling to native to get the Object pointer. This also fixes issues with the weaver not generating the marshaller correctly and also fixes the creation of the FScriptStruct which should have the second pointer point to the same pointer as the object.


Managed/UnrealSharp/UnrealSharp.SourceGenerators/InterfaceWrapperGenerator.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            })
+            .Where(x => x != null);
+        
+        context.RegisterSourceOutput(interfaceProvider, Execute!);
+    }
+
+    private static void Execute(SourceProductionContext context, INamedTypeSymbol interfaceSymbol)
+    {
+        /*
```
OlssonDev, line 36: Seems to be missing this?

OlssonDev:
How do I call a BP interface with these changes? :D

When doing reviews, I appreciate if the feature request comes with an example, so I can test it out easier and faster.

retroandchill:
It should just be the same as calling a C# or Native Interface. When the marshaller detects a non-implementing class was passed through it requests a wrapper object (which will create one the first time a wrapper is requested for this object, otherwise it will fetch the existing one). That wrapper implements the interface and by calling any method that is marked BlueprintEvent it will call the Blueprint, otherwise if the method is not implementable in Blueprints it will throw a NotImplementedException.

OlssonDev:
So how does it handle cases like this:

```
IMyInterface myInterface = (IMyInterface) UObjectRef;
myInterface.Function();
```

When I just have a UObject ref and I wanna call BP

retroandchill:
Unfortunately I don't believe you can cast like that. I think that's an omission on my part. I may need to add the ability to cast to an interface somehow.

OlssonDev:
So how do I call the BP implemented function in a case like that? Where I just have a UObject ref

retroandchill:
I currently have two ideas as to how this could be resolved:

Idea 1: 
Move the wrapper creation from the weaver to a source generator allowing the user to call it directly, while also adding an explicit conversion operation from UObject to the interface type (this may not work because I know C# has strange rules about conversion operators and interfaces).

Idea 2: 
Create an equivalent TScriptInterface struct that wraps the UObject and then source generate extension methods that are called on it for each interface that call to the managed interface methods.

retroandchill:
I just pushed a change that added new extension methods to handle converting between interfaces and UObjects. Now if you have a UObject and you need to call an interface method, you just call `uobject.AsInterface<IMyInterface>()`.

GameMill:
@retroandchill 
I think you will need to do the code standard fix for this PR as well.

"General fix is just brackets on new line and no use of auto in C++, just following the Unreal Engine code standard"

OlssonDev:
Seems like we need to support native interfaces too, not a whole lot of work. Do you want me to look into it? Some core things that needs to be changed in the data types for it to be clean.

retroandchill:
@OlssonDev If you want to take a look at it, that would be fine. Especially if you already have an idea of how you want to do it.

OlssonDev:
I'll start working on the native interfaces this weekend

OlssonDev:
Just committed a major rework on how we handle metadata, type info, type builders. It works on cropout, just gonna clean up the API from the refactor, then it's ready to be committed.

PR #471, opened 2025-07-05
Added support for the marshalling of FFieldPath and TFieldPath

retroandchill:
TFieldPath is a viable type for UPROPERTIES, used in a few places, most notably in FGameplayAttribute. To aid those trying to use this plugin with the GAS, having this type correctly marshalled and unmarshalled from struct types is imperative.


OlssonDev:
Awesome! :D I know this is requested for the GAS support. 

I just ran the formatter for some files and fixed up the binds on the C# side for FRandomStream

Thank you for the PR

PR #474, opened 2025-07-06
fix CsExporter check cache

RPG3D:
1. when glue folder deleted, invalid cache
2. add log when PublishSolution, it tooks a long time


OlssonDev:
Thank you for the PR!

PR #476, opened 2025-07-10
Improved interop between Native and C# when using Unreal Structs.

### LINKED ISSUE #475: [Feature] Overhaul how UStructs are handled in Managed Code
# Brief
The way USTRUCTS are marshalled and unmarshalled in C# is not ideal and can lead to memory loss and resource leaks in specific circumstances, and thus will need to change.

# Problem
The way C# handles structs is not the same as how they are handled in C++, meaning certain capabilities are absent that can lead to issues when interoping between managed and native code. The primary driver of this is that C# structs do not have the ability to have user-defined copy and destruction operations. For structs that consist solely of UPROPERTIES, this is a non-issue as all those types have marshalling and unmarshalling that respect lifetime (although `FFieldPath` and `TSoftObjectPtr` may warrant a memory safety investigation due to their usage of an `UnmanagedArray`). 

However, several struct types have additional properties not exposed to Blueprints (and thus not exposed to C#), which can lead to issues when a struct is taken into managed space and later passed back to native code, resulting in data loss. The main instance of this would be the various handle types in the Gameplay Ability System (GAS).

For example, the type `FGameplayEffectSpecHandle` wraps a non-exposed property of `TSharedPtr<FGameplayEffectSpec>` with `FGameplayEffectSpec` being a polymorphic type. Currently, this smart pointer is just lost when copied into C#; however, even if it were marshalled, it would not be able to properly update the reference counter, as C# structs lack any kind of inherent lifetime management.

The type is defined internally as such:
```C++
USTRUCT(BlueprintType)
struct FGameplayEffectSpecHandle
{
	GENERATED_USTRUCT_BODY()

	UE_API FGameplayEffectSpecHandle();
	UE_API FGameplayEffectSpecHandle(FGameplayEffectSpec* DataPtr);

	/** Internal pointer to effect spec */
	TSharedPtr<FGameplayEffectSpec>	Data;

       ...
};
```

On the flip side, C# defined USTRUCTS are also able to define fields that are not UPROPERTIES, which will not be marshalled and unmarshalled when interoping as well, leading to similar data loss issues. This may be used in an instance where a USTRUCT is being used as a lightweight handle to expose functionality in a 3rd party library that needs a class defined in that assembly. 

We could, for instance, have something similar to `FGameplayEffectSpecHandle` defined:
```C#
[UStruct(StructProperties.BlueprintType)]
public partial struct FMyManagedHandle
{
    // This can't be a UObject
    public IMyManagedType ManagedObject { get; }

    public FMyManagedHandle(IMyManagedType managedObject)
    {
        ManagedObject = managedObject;
    }
}
```

This type would also have marshalling issues, leading to that object reference being dropped when used in Blueprints. That object may also end up being garbage collected if it is being used in an Event Graph and a latent node is called, invalidating the entire process.

# Proposed Solution

# Native to C# Marshalling
We need a way to detect if a native USTRUCT that is being translated to C# contains non-UPROPERTIES. That may require additional header parsing. As far as I know, the UnrealHeaderTool (UHT) does not generate a full AST but rather hunts for specific tokens to find properties, so we may need a third-party solution such as [CppAst.NET](https://github.com/xoofx/CppAst.NET) to get a parse of the struct. We will also need to look for cases where it inherits from non-USTRUCT types (such as `FGCObject`) or has a user-defined destructor or copy constructor/assignment operation. If this is detected, the way the native wrapper is generated will need to change. In these cases, there will need to be a managed type that handles the struct lifetime. It would likely look something like this.

```C#
public sealed class NativeStructHandle : IDisposable 
{
    private readonly IntPtr _nativeStructPtr;

    private readonly IntPtr _nativeScriptStruct;

    private bool _disposed = false;

    public IntPtr NativeStructPtr 
    { 
        get 
        {
             if (_disposed) 
            {
                throw new InvalidOperationException("Native struct has already been disposed");
            }
           
            return _nativeStructPtr;
        }  
    }
    

    public NativeStructHandle(IntPtr nativeScriptStruct) 
    {
        _nativeScriptStruct = nativeScriptStruct;
        _nativeStructPtr= // Interop call to allocate and initialize
    }

    public NativeStructHandle(IntPtr existingStruct, IntPtr nativeScriptStruct) 
    {
        _nativeScriptStruct = nativeScriptStruct;
        _nativeStructPtr= // Call to allocate and copy the provided struct into the new one
    }

    ~NativeStructHandle() 
    {
        Dispose();
    }

    public void Dispose()
    {
        if (_disposed)
        {
            return;
        }
        
        // Native call to call the destructor and deallocate the memory
        _disposed = true;
    }

    public void CopyStruct(IntPtr otherStruct)
    {
        // Call to copy the struct in native code
    }
}
```
This could further be optimized by using various low-level memory optimizations such as small buffer optimization.

The generated struct would then look something like this:
```C#
[UStruct]
public partial struct FStructWithNativeMembers : IDisposable
{
    private NativeStructHandle? _handle;
    private static readonly IntPtr NativeScriptStructPointer;

    public int ExposedStructProperty 
    {
        get
        {
            // Native interop
        }
        set
        {
            // Native interop
        }
    }

    public FStructWithNativeMembers(IntPtr memory)
    {
        _handle = new NativeScriptStructPtr(memory, NativeScriptStructPtr);
    }

    public void Dispose()
    {
        _handle?.Dispose();
    }
}
```

Obviously, this would change some of the semantics as to how these structs work as these are not just references to an existing struct rather than a handle to the struct itself.

We may also want to generate additional struct alternatives for cases where the user wants to bind a struct by reference or const reference to cut back on the cost of making a copy, that could theoretically be handled something like this:
```C#
// Acts the same as `FStructWithNativeMembers&` in native code
[UStruct]
public partial ref struct FRefStructWithNativeMembers
{
    private NativeStructHandle? _handle;

    public int ExposedStructProperty 
    {
        get
        {
            // Native interop
        }
        set
        {
            // Native interop
        }
    }

    public FStructWithNativeMembers(IntPtr memory)
    {
        _handle = memory;
    }
}

// Acts the same as `const FStructWithNativeMembers&` in native code
[UStruct]
public partial readonly ref struct FReadOnlyRefStructWithNativeMembers
{
    private NativeStructHandle? _handle;

    public int ExposedStructProperty 
    {
        get
        {
            // Native interop
        }
    }

    public FStructWithNativeMembers(IntPtr memory)
    {
        _handle = memory;
    }
}
```

These structs would operate similar to how `Span<T>` and `ReadOnlySpan<T>` operate placing restrictions as to how these structs can be used (i.e. no boxing and no usage with `await` or `yield return`. For more information as to what these rules are see [Microsoft's documentation on ref structs](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/ref-struct).

## C# to Native Marshalling
Handling marshalling from C# to native and back again should be hypothetically simpler than the other way around as C# gives more tools to know the size and layout of structs ahead of time allowing for easier generation of marshalling code. For unmanaged types it would be as simple as adding custom properties to the struct that are not exposed to Blueprint and simply hold data of the same size as the C# type, the data is then copied in and out of native memory with no data loss.

However, when the property points to a managed type, this is where things get a bit more complicated. The approach will likely need to be one where a GCHandle is created and managed from the native C++ code. This could easily be handled by creating a custom property that wraps around a type that simply manages the memory through some kind of reference counting system.

It could be set up something like this:
```C++
class FManagedObjectRepository 
{
    private:
        FManagedObjectRepository ();
        ~FManagedObjectRepository ();

    public:
        static FManagedObjectRepository &Get();

        void AddReference(uint64 ManagedPointer);
        void RemoveReference(uint64 ManagedPointer);

    private:
        TMap<uint64, uint32> ReferenceCounts;
}

class FManagedPointerHandle
{
    public:
       FManagedPointerHandle() = default;

        explicit FManagedPointerHandle(uint64 InManagedPointer) noexcept : ManagedPointer(InManagedPointer)
        {
            if (ManagedPointer != 0) 
            {
                FManagedObjectRepository::Get().AddReference(ManagedPointer);
            }
        }

        FManagedPointerHandle(const FManagedPointerHandle& Other) noexcept : ManagedPointer(Other.ManagedPointer)
        {
            if (ManagedPointer != 0)
           {
                FManagedObjectRepository::Get().AddReference(ManagedPointer);
           }
        }

        FManagedPointerHandle(FManagedPointerHandle&& Other) noexcept : ManagedPointer(Other.ManagedPointer)
        {
            Other.ManagedPointer = 0;
        }

       ~FManagedPointerHandle(uint64 ManagedPointer)
       {
           if (ManagedPointer != 0)
           {
                FManagedObjectRepository::Get().RemoveReference(ManagedPointer);
           }
       }

       FManagedPointerHandle& operator=(const FManagedPointerHandle& Other) noexcept
       {
            if (ManagedPointer == Other.ManagedPointer)
            {
                return *this;
            }

            if (ManagedPointer != 0)
            {
                FManagedObjectRepository::Get().RemoveReference(ManagedPointer);
            }

            ManagedPointer = Other.ManagedPointer;

            if (ManagedPointer != 0) 
            {
                FManagedObjectRepository::Get().AddReference(ManagedPointer);
            }

            return *this;
       }

        FManagedPointerHandle& operator=(FManagedPointerHandle& Other) noexcept
       {
            ManagedPointer = Other.ManagedPointer;
            Other.ManagedPointer = 0;
       }

      

    private: 
        uint64 ManagedPointer = 0;
}
```

Obviously, the real implementation would likely need to be more complex, as C# can move memory around but this is the basic idea. This type would then be exposed as a property on the struct, and the managed object would then be marshalled into it. For the case of a managed struct type it would need to be boxed before being placed into native handling.

# Conclusion
There needs to be a change in how structs are handled under the hood. I plan to draft up a sample implementation on [my fork of the repo](https://github.com/retroandchill/UnrealSharp) in the coming days to see how viable this approach will be, and if any changes will be necessary

retroandchill:
Closes #475 

This PR improves how structs are handled in code.

The first change is that C# defined UStructs are now allowed to have fields of non-UProperty types, which will be marshalled in one of two ways.

Consider the following struct:
```C#
[UStruct]
public struct FStructWithManagedProperty
{
    [UProperty]
    public string NativeProperty;

    public int UnmarkedProperty;

    public ILookup<int, string> ManagedProperty;

    public Vector3 UnmanagedStruct;
}
```

In this instance, `NativeProperty` is marked as a UProperty and is thus marshalled as usual. The same goes for `UnmarkedProperty` as it is a marshallable type. The interesting changes come with `ManagedProperty` and `UnmanagedStruct`, which each represent the two ways marshalling is done. 

Looking at `ManagedProperty`, what we do is it is represented as a property of the type `FSharedGCHandle`, which essentially wraps a shared pointer to a `GCHandle`, meaning that once the reference counter hits 0, the handle is released and the GC is free to collect the object. This helps both keep the data intact when the struct is passed from Blueprints back to C#, and ensure the object isn't collected while it is still needed. This is the approach for all managed types, meaning any reference type or any struct that contains a reference type either directly or through a nested struct field. This does mean that a managed struct will end up being boxed.

For `UnmanagedStruct` instead, it is marshalled to the struct `FUnmanagedDataStore`, which is implemented as a wrapper around a `TVariant` object for either a static buffer or a `TSharedPointer` to an allocated chunk of memory. Which one gets used depends on the size of the C# type. If the data can fit in the buffer, that is what is used; otherwise, the space will be allocated and stored there.

As for marshalling Native to Managed, the approach is a bit more straightforward. The new star of the show is `NativeStructHandle`, which is intended to be used on structs that have non-UPROPERTIES and are not trivially destructible. This handle will ensure that any resources are not leaked by either releasing on GC or when the handle is explicitly disposed.

To aid in thi,s the UHT generator has been altered to take in JSON files to determine what types get special struct marshalling. An example of this is in the Config directory, which has the formerly hard-coded mappings. It contains two main sections. The first concerns blittable types, where the user can specify the name of the type and optionally the name of the managed class. The second section is for the types that have extra UPROPERTIES, which have the name and the `HasDestructor` flag, which will switch between using a simple byte-array as was done before or a `NativeStructHandle` to serve as the buffer for native code.

When using a struct that has a `NativeStructHandle` you can always ignore any management tasks and let the finalizer handle it, however if you want a more deterministic lifecycle management the structs that contain one of these handles now implement IDipsosable and can be manually disposed of with either a calling `Dispose()` directly or via a `using` statement.

Note that after an invocation of dispose, the handle is invalid and attempting to access any of the properties or marshalling the struct back to Blueprints will result in an exception.


retroandchill:
Added updates to how the handles are created. The allocated handle now uses small buffer optimization for storage, so any struct under 64 bytes will only require one allocation for the handle instead of one for the handle and another for the struct itself. I also added default constructors to structs that use native wrapping so that when they are initialized using new(), they are pre-populated with the default value of the struct as opposed to handling it lazily. That can still be skipped using `default` however.

OlssonDev (commented):
It looks great! :D Awesome stuff!

General fix is just brackets on new line and no use of auto in C++, just following the Unreal Engine code standard

retroandchill:
> It looks great! :D Awesome stuff!
> 
> General fix is just brackets on new line and no use of auto in C++, just following the Unreal Engine code standard

@OlssonDev 
I believe I have gotten all the problem spots in the PR.

OlssonDev:
Thank you for the great contributions the last weeks! Really awesome :) Thank you again <3

PR #479, opened 2025-07-13
Added support for TStrongObjectPtr in C# code

retroandchill:
Added a C# equivalent to the C++ only class TStrongObjectPtr which is a way to storing a single UObject in a class that is not part of the UProperty system. This provides the benefit of allowing for regular C# classes to hold a reference to a UObject without fear of it getting garbage collected by the engine. This type is able to be used as such:

```C#
public class MyRegularClass 
{
    private readonly TStrongObjectPtr<UMyObjectType> _myObject;

    public MyRegularClass(UMyObjectType myObject) 
    {
        _myObject = myObject; // Implicit conversion operator
    }
}
```

This will ensure that so long as the TStrongObjectPtr is in memory so will the object it wraps. Alternatively the user can also manually manage the lifetime of the object by using the Dispose pattern, which is recommended since it provided more deterministic guarentees about object lifetime, which is helpful in avoiding fatal engine crashes from things like not allowing Actors to be collected when the world is.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/Interop/TStrongObjectPtrExporter.cs
+++ b/Managed/UnrealSharp/UnrealSharp/Interop/TStrongObjectPtrExporter.cs
@@ -1,0 +1,10 @@
+﻿using UnrealSharp.Binds;
+
+namespace UnrealSharp.Interop;
+
+[NativeCallbacks]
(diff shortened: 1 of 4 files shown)


OlssonDev:
Great PR! :D Thank you for the contribution!

Issue #98, opened 2024-04-01
Support for records


thomasfn:
Records are a nice way of reducing boilerplate for struct-like types. It would be handy if `UStruct` was smart enough to work on records and have the weaver or sourcegen take care of all the `UProperty` boilerplate with sensible flags.

For example, the following record...
```C#
[UStruct]
public record struct BezierPoint(Vector3 Position, double Tilt);
```

...could be expanded to...
```C#
[UStruct]
public struct BezierPoint
{
    [UProperty(PropertyFlags.BlueprintReadWrite | PropertyFlags.EditAnywhere)]
    public Vector3 Position { get; set; }

    [UProperty(PropertyFlags.BlueprintReadWrite | PropertyFlags.EditAnywhere)]
    public double Tilt { get; set; }
}
```

Unfortunately class records (which might be appropiate for `UDataAsset` subclasses) can't extend anything except `System.Object` and other class records so we can't give `UClass` the same treatment.

PR #480, opened 2025-07-13
Allow for properties on UStructs instead of just fields.

retroandchill:
This ties into #98 by helping allow devs to use record structs for structs. To do this, I had to remove the restriction for not allowing any properties on UStructs and instead make the requirement that for a property to be reflected, it must still be reflected via its backing field. The same goes for using `record struct` classes.

For that, you're able to declare it as such:
```C#
[UStruct]
public record struct FMyDemoStruct([field: UProperty(PropertyFlags.EditAnywhere)] int MyInt, 
                                   [field: UProperty(PropertyFlags.EditAnywhere)] string MyString);
```

What happens under the hood is that the weaver will check if a field has a name that matches the compiler-generated backing field name, and if so, generate a property with the name of what the property would be based on the name of the backing field that is mapped to that initial field. 

I have also changed the glue generator to utilize properties rather than fields just to stay more in-line with what standard C# practices are.

### DIFF
--- a/Config/Default.UnrealSharpTypes.json
+++ b/Config/Default.UnrealSharpTypes.json
@@ -1,0 +1,102 @@
+{
+    "Structs": {
+        "BlittableTypes": [
+            {
+                "Name": "EStreamingSourcePriority"
+            },
+            {
+                "Name": "ETriggerEvent"
+            },
+            {
+                "Name": "FVector",
+                "ManagedType": "UnrealSharp.CoreUObject.FVector"
+            },
+            {
+                "Name": "FVector2D",
+                "ManagedType": "UnrealSharp.CoreUObject.FVector2D"
+            },
+            {
+                "Name": "FVector_NetQuantize",
+                "ManagedType": "UnrealSharp.CoreUObject.FVector"
+            },
+            {
+                "Name": "FVector_NetQuantize10",
+                "ManagedType": "UnrealSharp.CoreUObject.FVector"
+            },
+            {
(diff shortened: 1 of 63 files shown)


OlssonDev:
This is great! Although I think we want to keep the native members as fields. For example we can no longer do this if they're properties:

`transform.Scale.Z = 1.0f;`

Then the user gets to specify if they want properties for their own structs instead.

retroandchill:
@OlssonDev I reverted the usage of properties on generated structs. I also fixed some of the strange history from reverting the merges.

OlssonDev:
Merging latest into this PR throws this error:

`CSWorldSubsystem.generated.cs(366, 30): [CS9244] The type 'FSubsystemCollectionBaseRef' may not be a ref struct or a type parameter allowing ref structs in order to use it as parameter 'T' in the generic type or method 'StructMarshaller<T>' `

retroandchill:
> Merging latest into this PR throws this error:
> 
> `CSWorldSubsystem.generated.cs(366, 30): [CS9244] The type 'FSubsystemCollectionBaseRef' may not be a ref struct or a type parameter allowing ref structs in order to use it as parameter 'T' in the generic type or method 'StructMarshaller<T>' `

I think this branch got corrupted somehow because it's trying to revert a bunch of other changes. I'll take a look at it later today.

retroandchill:
@OlssonDev I just force pushed fixes to revert the corrupted history. It should only show 2 changed files now

PR #482, opened 2025-07-16
Fixed TSoftClassPtr<T>

blizzzrdda:
`softClassProperty.Class.GetFullManagedName()` translated T as UClass, which is supposed to be an UObject.


OlssonDev:
Thank you for the fix! 😄

PR #484, opened 2025-07-19
Fixed an issue where if a project plugin lacked a config directory the UHT would crash

retroandchill:
This fixed an issue that was reported where having plugins in the project without a Config directory the project could not be built.


OlssonDev:
Thank you for the fix :D

PR #485, opened 2025-07-19
Allow users to define analyzers in their project code

retroandchill:
Currently, if the project encounters an analyzer during the initial build or during hot reload of the plugin will display an error. This fixes it by ensuring any .csproj files in the project that have `IsRoslynComponent` set to `true` are ignored by the weaver and hot reloader as they are build-only artifacts.


GameMill:
I'd like to propose a change here. 
Rather than checking for IsRoslynComponent set to true, consider switching to ExcludeFromWeaver set to true.
This would give more flexibility by allowing other project types to be excluded from the weaver  and hot reload functionality as well if required.

OlssonDev:
Awesome PR! Thank you <3

PR #486, opened 2025-07-19
Make UE collections implement the read-only versions of the interfaces

retroandchill:
Currently, TArray, TSet, and TMap only implement the read/write interfaces for their collection types. However, if you look at the definitions for List, HashSet, and Dictionary, they also implement the read-only versions of those collections. This PR fixes that by ensuring the three collections also implement the read-only versions.


OlssonDev:
Ohh! Nice :D Thank you for the PR!

PR #487, opened 2025-07-19
Enable UObjects to have get only properties

retroandchill:
Right now, the weaver crashes if a UObject property does not have a setter defined. However, a user may want to communicate that a property is only retrievable in C#, and is set by the editor or during construction. This PR rectifies that by skipping over set method generation for properties that lack a pre-defined set method.


OlssonDev:
Great change! Thank you for the PR :)

PR #488, opened 2025-07-19
Updated subsystems so there is a way to request dependencies in C#

retroandchill:
In C++ subystems are able to declare dependencies through the `InitializeDependency` method on `FSubsystemCollectionBase`, however since this type is not exposed to Blueprints there is no way to handle this same thing in C#. To rectify this, I created `FSubsystemCollectionBaseRef` as a wrapper struct that holds a pointer to the underlying `FSubsystemCollectionBase` object. This is then reflected in C# by the type having a single property of type IntPtr and being marked as Blittable. The struct is also defined as a readonly ref struct to ensure that it cannot escape the context of the initialize method that called it. Additional support code was also modified to add `allows ref struct` so that this struct can be defined this way. 

The struct has two distinct method overload sets. `InitializeDependency` and `InitializeRequiredDependency`, with the former having the capacity to return null and the latter throwing an `InvalidOperationException` if it cannot find a dependency object.

The usage would look something like this:
```C#
[UClass]
public class MySubsystem : UCSGameInstanceSubsystem 
{
    [UProperty]
    private UDependentSubsystem Dependency { get; set; }

    protected override void Initialize(FSubsystemCollectionBaseRef collection)
    {
        Dependency = collection.InitializeRequiredDependency<UDependentSubsystem>();
    }
}
```

### DIFF
--- a/Config/Default.UnrealSharpTypes.json
+++ b/Config/Default.UnrealSharpTypes.json
@@ -70,6 +70,10 @@
             {
                 "Name": "FRandomStream",
                 "ManagedType": "UnrealSharp.CoreUObject.FRandomStream"
+            },
+            {
+                "Name": "FSubsystemCollectionBaseRef",
+                "ManagedType": "UnrealSharp.UnrealSharpCore.FSubsystemCollectionBaseRef"
             }
         ],
         "NativelyTranslatableTypes": [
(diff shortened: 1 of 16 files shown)


OlssonDev:
Awesome, didnt know about this feature in subsystems! :D Thank you for the PR!

PR #490, opened 2025-07-25
Fix: Editor-only include caused UnrealSharp.Plugins.dll to fail compilation in packaged builds.

GameMill:
Fixes an issue where UnrealSharp.Plugins.dll would silently fail to compile during packaged builds, resulting in the following error:

"LogUnrealSharp: Error: UnrealSharp.Plugins.dll does not exist at: /Binaries/Managed/UnrealSharp.Plugins.dll"


OlssonDev:
Thank you for the fix!

PR #493, opened 2025-07-27
Fixed generation of WITH_EDITOR properties

retroandchill:
Currently U# generates glue code that is editor only using the `WITH_EDITOR` preprocessor directive, but that is never set to true, meaning that code is never compiled. This PR replaces that with the `!PACKAGE` directive that is used elsewhere in the code.


OlssonDev:
I think the better approach is to remove PACKAGE and just do WITH_EDITOR, to closely resemble the c++ API. 

I could do that change to this commit. Awesome fix <3

retroandchill:
> I think the better approach is to remove PACKAGE and just do WITH_EDITOR, to closely resemble the c++ API.
> 
> I could do that change to this commit. Awesome fix <3

I think the only thing we need to ensure is that WITH_EDITOR is enabled by default when editing a user's solution that way those blocks are not greyed out by VS/Rider.

OlssonDev:
With these changes to the glue generator, the generator no longer take into account any BlueprintGetter/BlueprintSetter. `RootComponent` for example is now protected with these changes, since it ignores its public BlueprintGetter

OlssonDev:
Awesome thank you for the fix! :D

PR #499, opened 2025-08-02
Include CSMetaDataUtils.h

hez2010:
Otherwise it would result in a build error complaining about `FCSMetaDataUtils` not found.


OlssonDev:
Thank you! This compiler error was not on my end, are you using unity build?

hez2010:
> This compiler error was not on my end, are you using unity build?

I noticed this when I pulled the plugin to a brand new UE5.6 project.

PR #502, opened 2025-08-03
Support initialize readonly TArray, TSet and TMap

### LINKED ISSUE #469: [Issue] Can't init some properties in C#
This is related to #467, there seems to be no way of initializing these properties in C#, especially when working with subobjects.

In C++ I have a UActorComponent defined like this:

```cpp
#pragma once

#include "CoreMinimal.h"
#include "Components/ActorComponent.h"
#include "MyCustomComponent.generated.h"


UCLASS(ClassGroup=(Custom), meta=(BlueprintSpawnableComponent))
class CROPOUTSAMPLEPROJECT_API UMyCustomComponent : public UActorComponent
{
	GENERATED_BODY()
	
public:
	UPROPERTY(EditDefaultsOnly, BlueprintReadOnly)
	TArray<FString> Names;

protected:
	UPROPERTY(EditDefaultsOnly, BlueprintReadOnly)
	TArray<FString> ProtectedNames;
};
```

Which generates the following in C#

```csharp
[UClass, GeneratedType("MyCustomComponent", "UnrealSharp.CropoutSampleProject.MyCustomComponent")]
public partial class UMyCustomComponent : UnrealSharp.Engine.UActorComponent
{
    //...

    public System.Collections.Generic.IReadOnlyList<string> Names
    {
        get
        {
             //....
        }
    }

    //...

    protected System.Collections.Generic.IReadOnlyList<string> ProtectedNames
    {
        get
        {
             //....
        }
    }
}
```

Both of these are `IReadOnlyList` so they don't have the Add() method. And because they don't have a setter I can't initialize them by setting a new list either. Also, even if they had a setter, it's possible that #445 will also come into play. When I try to init them in C#, I get `The property .... has no setter`.

```csharp
[UClass]
public class AActorTest: AActor
{
    public AActorTest()
    {
        // Error: The property 'UnrealSharp.CropoutSampleProject.UMyCustomComponent.Names' has no setter
        MyCustomComponent.Names = new List<string> {"A", "B", "C"};
    }
    
    [UProperty(DefaultComponent = true)]
    public UMyCustomComponentCsharp MyCustomComponent { get; init; }
}

[UClass]
public class UMyCustomComponentCsharp: UMyCustomComponent
{
    public UMyCustomComponentCsharp()
    {
        // Error: The property 'UnrealSharp.CropoutSampleProject.UMyCustomComponent.Names' has no setter
        Names = new List<string> {"A", "B", "C"};
        ProtectedNames = new List<string> {"D", "E"};
    }
}
```

I'm unsure how this should be handled, because if the properties where defined with `{get; init;}` then we can initialize the array in the subclass `UMyCustomComponentCsharp`, but It doesn't seem to be posible to init it in `AActorTest`. A setter seems to be required for this to work. It could be fixed in this case for TArrays if it generated an `IList` instead, but a similar issue happens for other variable like:

```cpp
public:
	UPROPERTY(EditDefaultsOnly, BlueprintReadOnly)
	FString Name;
```

Where `Name` is generated with a `{get; init;}` in C#, but then we can't initialize the value in an Actor, only in a component subclass.

hez2010:
Generates `{ get; init; }` instead of `{ get; }` to allow readonly collections being initialized in constructors.

Fixes #469

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/Array.cs
+++ b/Managed/UnrealSharp/UnrealSharp/Array.cs
@@ -160,8 +160,13 @@ public class TArray<T> : UnrealArrayBase<T>, IList<T>, IReadOnlyList<T>
 public class ArrayMarshaller<T>(IntPtr nativeProperty, MarshallingDelegates<T>.ToNative toNative, MarshallingDelegates<T>.FromNative fromNative)
 {
     private TArray<T>? _arrayWrapper;
-    
+
     public void ToNative(IntPtr nativeBuffer, int arrayIndex, IList<T> obj)
+    {
+        ToNative(nativeBuffer, obj);
+    }
+
+    public void ToNative(IntPtr nativeBuffer, IList<T> obj)
     {
         unsafe
         {
--- a/Managed/UnrealSharp/UnrealSharp/ArrayReadOnly.cs
+++ b/Managed/UnrealSharp/UnrealSharp/ArrayReadOnly.cs
@@ -1,5 +1,6 @@
 using UnrealSharp.Core;
 using UnrealSharp.Core.Marshallers;
+using UnrealSharp.Interop;
 
 namespace UnrealSharp;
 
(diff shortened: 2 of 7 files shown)


OlssonDev:
Thank you for the fix and PR! :D

PR #509, opened 2025-08-09
Made adjustments to how Metadata tags work

retroandchill:
This PR adds some additional metadata tags (mainly ones pertaining to Gameplay Tags), while also expanding the usage rules for these Attributes a bit. 

- All UProperty tags are now allowed on fields as there are cases where these tags are beneficial on UStructs
- A select few that are valid on method parameters are also allowed there

This PR also makes adjustments to ensure that parameter metadata is properly loaded by U# at runtime.

There are also a few assorted bug fixes that I found when testing.


OlssonDev:
Awesome! Thank you for the PR :D

PR #514, opened 2025-08-11
UE_LOGFMT not found and purge metadata GetMapForObject fix 

abordiuh:
Replace UE_LOGFMT (no include header) to UE_LOG, fix GetMapForObject left value error


abordiuh:
Ok, ignore that, I see you guys just already solved these issues

PR #516, opened 2025-08-12
Various fixes and improvements

hez2010:
- Add the exception message and stacktrace to the log when the managed counterpart was failed to create to improve the diagnosability when an exception was thrown from the constructor
- Replace `Marshal.GetDelegateForFunctionPointer` and `Marshal.GetFunctionPointerForDelegate` with function pointers to reduce the overhead
- Properly generates the setter, and skip the initialization for `ldnull` to avoid marshalling a null value for types like `TArray` and etc. A field with reference type should already have `null` as the initial value by default (a follow up to #512)
- Tweak the global.json to allow using a prerelease version of .NET SDK
- Only purge metadata when it's available to avoid build errors during packaging with Shipping configuration
- Override the `OutputPath` to use the intermediate path so that assemblies won't be locked by the unreal editor which can lead to failure to package the game

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp.Binds/BindsManager.cs
+++ b/Managed/UnrealSharp/UnrealSharp.Binds/BindsManager.cs
@@ -1,38 +1,38 @@
-using System.Runtime.InteropServices;
-
-namespace UnrealSharp.Binds;
-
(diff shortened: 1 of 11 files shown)


OlssonDev:
Awesome fixes! Thank you for the PR :D

PR #520, opened 2025-08-17
Incremental source generators

hez2010:
Implement all source generators to use incremental source generators to improve edit-time performance.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp.EditorSourceGenerators/UnrealSharp.EditorSourceGenerators/ClassFilePathGenerator.cs
+++ b/Managed/UnrealSharp/UnrealSharp.EditorSourceGenerators/UnrealSharp.EditorSourceGenerators/ClassFilePathGenerator.cs
@@ -14,99 +14,8 @@ struct AssemblyClassInfo(string fullName, string filePath)
 }
 
 [Generator]
-public class ClassFilePathGenerator : ISourceGenerator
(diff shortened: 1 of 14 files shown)


OlssonDev:
Awesome :D 

When I try to compile it just says `UnrealLogger.cs(12,30): Error CS0117 : 'FMsgExporter' does not contain a definition for 'CallLog'`

hez2010:
> When I try to compile it just says `UnrealLogger.cs(12,30): Error CS0117 : 'FMsgExporter' does not contain a definition for 'CallLog'`

Maybe you need to regenerate the C++ project and trigger a fresh build? I cannot reproduce it in my local environment.

`CallLog` is generated by the source generator correctly:
image

GameMill:
@OlssonDev I just compiled this on Cropout and didn’t encounter this issue.

OlssonDev:
`UnrealSharp.SourceGenerators.dll' references version '4.14.0.0' of the compiler, which is newer than the currently running version '4.13.0.0'.`

I'm getting this with just .NET 9.0.100 installed, which is causing the error in FMsgExporter

hez2010:
Okay it seems that I need to downgrade the compiler dependency.

hez2010:
@OlssonDev Please take another look

OlssonDev:
Thank you for the fix :D

PR #523, opened 2025-08-20
Prevent "Handle is not initialized" error when UObject lacks interface implementation

GameMill:
This update adds a guard clause to prevent a runtime error when AsInterface is called on a UObject that does not implement the expected interface. Previously, wrapperHandle could return an invalid handle (IntPtr.Zero), leading to a "Handle is not initialized" exception.

image


OlssonDev:
Thank you for the PR :D

PR #524, opened 2025-08-22
Fix : Can't get value of function parameter which is ReferenceParam.

ccSugarCole:
When I mark a function parameter as 'ref' in c#, and use blueprint or C++ to call this function, the parameter will be marked as PropertyFlags.OutParm and PropertyFlags.ReferenceParm together.

The according to the PropertyFlags.OutParm, the value of parameter can't be got.

### DIFF
--- a/Managed/UnrealSharpPrograms/UnrealSharpWeaver/TypeProcessors/FunctionProcessor.cs
+++ b/Managed/UnrealSharpPrograms/UnrealSharpWeaver/TypeProcessors/FunctionProcessor.cs
@@ -185,7 +185,7 @@ public static class FunctionProcessor
             
             param.PropertyDataType.PrepareForRewrite(type, param, methodToCall);
 
-            if (param.PropertyFlags.HasFlag(PropertyFlags.OutParm))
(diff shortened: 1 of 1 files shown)


ccSugarCole:
image
I want to us c# script in here, so I must use 'ref for PropertyFlags.ReferenceParm to bind function here.

OlssonDev:
So without the change to the weaver, you can't call functions from the ABP?

ccSugarCole:
> So without the change to the weaver, you can't call functions from the ABP?

yes. I must to use ‘ref’.
Function which need to bind in here, will compare to a PrototypeFunction by UFunction::IsSignatureCompatibleWith, and the PrototypeFunction is AnimExecutionContextLibrary.Prototype_ThreadSafeAnimUpdateCall.
It's necessary to mark PropertyFlags.ReferenceParm for UFunction::IsSignatureCompatibleWith to return true.

image
image

I think that only 'out' param don't need Construct, but 'ref' param should do.

OlssonDev:
Thank you for the fix :D

PR #525, opened 2025-08-22
Updated assembly loading to load from the actual path rather than a stream

retroandchill:
This PR makes changes to how assembly loading is done:

1. Assembly loading is now done from a path instead of from a stream
2. The IL Weaver output during hot reload is placed in an intermediate directory and then copied over after unloading

This also adds a new delegate for responding to when an assembly is unloaded for native code.


OlssonDev:
Looks good! :D 

Some conflicts from main that needs to be resolved before merge

retroandchill:
@OlssonDev Just merged and pushed up the fix

Unreal-3D:
load binary from stream makes it possible to load binary from UE's pak file system, its important for mobile and console in future, such as in-game hotfix binary

OlssonDev:
Why was this closed?

PR #526, opened 2025-08-23
Fix MulticastDelegate.Clear not working

hez2010:
The order of argument `NativeProperty` and `NativeDelegate` was wrong...


OlssonDev:
Thank you for the fix :D

PR #527, opened 2025-08-24
Fix package version

hez2010:
I accidently made a typo where the version of `Microsoft.CSharp` should be 4.7.0, instead of 4.13.0.
Also move `Directory.Package.props` to the parent directory so that it also applies to projects under `Source.`


hez2010:
Ping. This is a simple fix that sets the correct version for `Microsoft.CSharp`.

OlssonDev:
Sorry, been out sick all week. Not been able to be at the computer. Merged now!

PR #528, opened 2025-08-27
Refactor directory traversal logic to locate project root from plugin path

yajiedesign:
The current way of locating project directories assumes that plugins are placed in the Plugins directory. 
However, developers may need to further organize plugins using subdirectories. 
This PR changes the way project directories are located, looking upwards to the Plugins directory to support placing plugins in subdirectories.Refactor directory traversal logic to locate project root from plugin path.


OlssonDev:
```
        string projectDirectory = Factory.Session.ProjectDirectory;
        string pluginsDirectory = Path.Combine(projectDirectory, "Plugins");
```

Great catch :D But we can just do this instead! Less code, yippie!

PR #534, opened 2025-09-02
Gameplay attributes

blizzzrdda:
## Expose Gameplay Attributes with Static Accessor Pattern
Implemented a static accessor pattern for Gameplay Attributes similar to the existing `GameplayTags.cs` approach. This enhancement provides convenient compile-time access to gameplay attributes through static properties.

### Key Changes:
- Added static list of accessors for gameplay attributes in `GameplayAttributes.cs`
- Introduced UCSGameplayAttributeSubsystem (Engine Subsystem) to initialize attribute accessors at runtime
- Runtime initialization is required because FProperty references must be captured during engine startup

### Known Issue:
The current implementation introduces a dependency issue where UnrealSharpRuntimeGlue now depends on UnrealSharpCore due to attribute glue generator and  UCSGameplayAttributeSubsystem usage. I am not sure whether it can cause circular dependency in future

### DIFF
--- a/Source/UnrealSharpCore/Extensions/Subsystems/CSGameplayAttributeSubsystem.cpp
+++ b/Source/UnrealSharpCore/Extensions/Subsystems/CSGameplayAttributeSubsystem.cpp
@@ -1,26 +1,26 @@
 // Fill out your copyright notice in the Description page of Project Settings.
 
 
-#include "GameplayAttributeSubsystem.h"
+#include "CSGameplayAttributeSubsystem.h"
 
-void UGameplayAttributeSubsystem::Initialize(FSubsystemCollectionBase& Collection)
+void UCSGameplayAttributeSubsystem::Initialize(FSubsystemCollectionBase& Collection)
 {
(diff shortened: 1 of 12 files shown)


OlssonDev:
Great feature!

I'm thinking if this might be better as a plugin, since it involves GAS! We're gonna start a GAS plugin soon enough, but this can be a first step. 

Don't wanna clog the base plugin with big features like GAS, if someone doesn't use it.

retroandchill:
To add to this, I have a working implementation over on my fork from my attempts to prototype a CommonUI integration that allows you to generate the native glue code for any engine module (and any of its dependencies) and place them into your own plugin. That might be helpful for certain aspects of the GAS implementation, such as adding implicit conversions between float and FGameplayAttributeData.

OlssonDev:
https://github.com/UnrealSharp/UnrealSharpGAS/commit/28721f4ea7d5fadedff685e3a7e7df7c3e22d27e Added through the plugin instead! Thank you for the PR :D

PR #536, opened 2025-09-04
Implemented autogenerated Struct equality and arithmetic

FailMelon:
A lot of basic structs with primitive based field type currently don't have equality, this PR aims to resolve this by autogenerating equality interface on every valid struct. This also includes an exclusion list if we need to go into more detail on equality ourselves.

Also added basic arithmetic logic to supported structs as well.


OlssonDev:
Awesome feature! Thank you for the PR :D

PR #537, opened 2025-09-04
Added additional features when working with Structs

retroandchill:
This PR pulls in a couple of features related to structs that help with interop generation:

1. Adjusts the weaver such that all user-generated structs now implement from MarshalledStruct<T>. This is coupled with a source generator that, when a C# struct is marked as partial, generates a stub to indicate that the class implements this interface with all the interface methods marked as extern. This allows for generic constraints on UStructs to reliably use the MarshalledStruct<T> type as an option. The weaver can notice that a struct already implements this interface and uses the existing extern methods instead of creating its own, but that is strictly optional.
2. Adds basic support for customizing how the glue code for native structs is generated. This includes specifying the ability to use properties instead of fields and to mark the struct as read-only, allowing for effective immutability. Users can also make them as a record (although without the concise syntax currently).
3. The third is to add support for FInstancedStruct, which is a custom struct type that holds a pointer to a UScriptStruct object and some opaque memory. It is essentially a way to allow for struct-based polymorphism for small sets of data. This adds all the basic extensions for the type and also adds the ability to completely skip over the generation of glue code for structs to allow users to create their own custom marshalling code.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/Extensions/CoreUObject/Guid.cs
+++ b/Managed/UnrealSharp/UnrealSharp/Extensions/CoreUObject/Guid.cs
@@ -8,7 +8,7 @@ using System.Threading.Tasks;
 
 namespace UnrealSharp.CoreUObject;
 
-public partial struct FGuid : IEquatable<FGuid>
+public partial struct FGuid
 {
     public FGuid(Guid guid)
     {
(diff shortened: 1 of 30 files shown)


FailMelon:
MSBuild doesn't allow dynamic TFM switching sadly so this will still have same issue now just with .net 6 builds, from when i experimented it seemed to be fine building via UBT not sure what magic unreal does but if the user tries to build the sln it'll try to build the ubtplugin via normal MSBuild and fail.

retroandchill:
I can take that stuff out. I added it in because I needed to test the UBT stuff in isolation and it won't build from the sln

FailMelon:
Its fine probably for best we have it working latest first, i'm wondering what we can do about it though. We might have to start making branches for engine versions for stuff like this.

OlssonDev:
Cool stuff! :D Thank you for the PR!

PR #538, opened 2025-09-04
Added a new event for when an assembly is unloaded

retroandchill:
Adds a new event dispatcher to listen for when an assembly is unloaded so any native code that is holding onto managed handles has an opportunity to drop those handles.


OlssonDev:
Thank you for the PR! :D

PR #539, opened 2025-09-13
Added logic to allow an engine modules to be extracted out info a plugin

retroandchill:
This PR is a fairly straightforward. It allows for users to extract specific engine modules out into their own plugins for glue generation. This might come in handy for extending plugins that are not enabled by default (i.e. GAS). This will also help users who need to integrate with FAB plugins as well.


OlssonDev:
Awesome! :D Thank you!

PR #540, opened 2025-09-13
Added support for getting the managed types that are used for UScriptStruct and UEnum

retroandchill:
This PR allows users to take a UScriptStruct or UEnum object and get the managed type out of it. Unlike the way that UClass does it, these have the possibility of returning null, which would indicate that the user called it on a Blueprint Struct/Enum.

### DIFF
--- a/Source/UnrealSharpCore/CSManager.cpp
+++ b/Source/UnrealSharpCore/CSManager.cpp
@@ -597,7 +597,7 @@ UCSAssembly* UCSManager::FindOwningAssembly(UEnum* Enum)
 }
 
 
-UCSAssembly * UCSManager::FindOwningAssemblySlow(UField *Field)
+UCSAssembly* UCSManager::FindOwningAssemblySlow(UField *Field)
 {
     // Slow path for native classes. This runs once per new native class.
     const FCSFieldName ClassName = FCSFieldName(Field);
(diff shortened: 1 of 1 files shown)


Source/UnrealSharpCore/CSManager.cpp
```diff
@@ -536,24 +538,81 @@ UCSAssembly* UCSManager::FindOwningAssembly(UClass* Class)
...
+	
+    if (const ICSManagedTypeInterface* ManagedType = Cast<ICSManagedTypeInterface>(Struct); ManagedType != nullptr)
+    {
+        // Fast access to the owning assembly for managed types.
+        return ManagedType->GetOwningAssembly();
+    }
+
+    if (const UUserDefinedStruct* UserStruct = Cast<UUserDefinedStruct>(Struct); UserStruct != nullptr)
+    {
+        // This is a Blueprint Struct and we can't use it
```
OlssonDev, line 557: When will we get a Blueprint struct in this scenario?

retroandchill: If you have an instance of something like `FInstancedStruct` or you have a context where you get a `UScriptStruct` object directly.

OlssonDev:
Sorry for slow merge. Life happened :D 

Thank you for the PR!

PR #541, opened 2025-09-13
Various fixes to code generation

retroandchill:
This PR includes a number of fixes/improvements for the UBT, weaver, and source generators that improve the developer experience dramatically.

That list includes:

- Fixes involving generic parameters in generated glue code, including extension methods, and when a dynamic output param is an out parameter instead of a return value. Also, ensures any of the _Extension files are also added to the list of unchanged files during glue generation. 
- Improvements to diagnostic suppression of nullable UProperties to help reduce some false-positive warnings about uninitialized properties
- Fixes to async UFunctions, including fixing issues where the methods are uncallable across the assembly boundary, as well as adding support for ValueTask.
- Added support for the IsA method on TSubclassOf
- Added some Span support for FText and FName to allow for some optimizations
- Added the BindWidgetOptional attribute
- Added fixes for native getters and setters on properties, including a fix to not call invalid UFunctions on skeleton classes that was causing a crash
- Fixed some issues with interfaces with get/set methods that conflict with UProperties on native classes, done through explicit interface implementations
- Fixed some issues with the selection of marshallers selecting the wrong marshaller when there is a name conflict, by simply using StructMarshaller for everything now
- Fixed an issue with the return value being overwritten/ignored when UFunctions return to C++.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -14,5 +14,6 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="4.13.0" />
     <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageVersion Include="Microsoft.Build.Locator" Version="1.9.1" />
+    <PackageVersion Include="PolySharp" Version="1.15.0" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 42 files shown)


Directory.Packages.props
```diff
@@ -14,5 +14,6 @@
     <PackageVersion Include="Microsoft.CodeAnalysis.Workspaces.Common" Version="4.13.0" />
     <PackageVersion Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageVersion Include="Microsoft.Build.Locator" Version="1.9.1" />
+    <PackageVersion Include="PolySharp" Version="1.15.0" />
```
OlssonDev, line 17: What is this dependency used for?

hez2010: Use newer C# language feature in netstandard2.0 (basically introducing some internal types for C# compiler to recognize).
It's completely a private asset dependency so that it won't be exposed to users.

GameMill: Since UnrealSharp.SourceGenerators still compiles without it, I’m wondering if it’s actually being used. If not, we could think about removing it.

retroandchill: I think I added it without actually using it. We can take it out if it's not being used.

GameMill: @retroandchill Don’t forget to remove PolySharp from Directory.Packages.props too.

OlssonDev:
Thank you for the fixes! I did just a formatting pass :) Looks good!

PR #545, opened 2025-09-18
Added inverse transform math to FTransform

### LINKED ISSUE #543: FTransform missing Inverse methods
The C#-side `FTransform` struct has `TransformPosition` and `TransformVector` but not `InverseTransformPosition` and `InverseTransformVector`. There's also no way to invert the transform itself. You can cast it to an `FMatrix` but there's no way to invert that either.

The inverse options in `FTransform` are pretty critical for doing even basic and common transformation operations, e.g. moving from world space to component space.

Ideally these functions would be implemented in pure C# to avoid interop overhead.

FailMelon:
Fixes #543 provided by @thomasfn


OlssonDev:
Thank you for the PR :D

PR #546, opened 2025-09-19
Refactor TOptional support

FailMelon:
As discussed with Olsson this PR removes the need to rely on LanguageExt and refactors optional to be inside of UnrealSharp instead of UnrealSharp.Core. This was done due to the bloat that LanguageExt has a library which should not be forced onto user projects, this does currently remove some of features from the Option from that library. However should we need to we can add our own stuff to this if need be.


FailMelon:
Blocked by #541 will fix after merge

retroandchill:
I generally agree with this. I went with LanguageExt as an easy route to get the support in, but I understand that this library is very opinionated in how you operate. We don't need to do it for this merge, but after this gets merged and incremental source generation is complete, adding additional support for nullable value types as another option, as that's built directly into the library. As mentioned as well, we can also add the monads back in at a later point.

FailMelon:
Yeah nullable value type is something we should look to support

OlssonDev:
The other one is merged now!

OlssonDev:
Thank you for the PR :D

Issue #459, opened 2025-06-29
Incomplete property generated for Inkpot plugin


CuppaXanax:
While attempting to use the plugin [Inkpot](https://github.com/The-Chinese-Room/Inkpot) with U#, I hit the following error trying to compile the project glue generated for the plugin:

```msbuild
N:\dev\Games\Unreal\InkpotUnrealSharpBrk\Script\ProjectGlue\InkpotEditor\BlotterVariable.generated.cs(206,9): error CS0161: 'UBlotterVariable.DisplayValue.get': not all code paths return a value [N:\dev\Games\Unreal\InkpotUnrealSharpBrk\Script\ProjectGlue\ProjectGlue.csproj]
```

I can fix the build manually by patching the generated file:

```csharp
    public string DisplayValue
    {
        get
        {
            unsafe
            {
                byte* paramsBufferAllocation = stackalloc byte[GetDisplayValue_ParamsSize];
                nint paramsBuffer = (nint) paramsBufferAllocation;
                
                UObjectExporter.CallInvokeNativeFunctionOutParms(NativeObject, GetDisplayValue_NativeFunction, paramsBuffer, paramsBuffer + GetDisplayValue_ReturnValue_Offset);
                
                FText returnValue;
                returnValue = TextMarshaller.FromNative(IntPtr.Add(paramsBuffer, GetDisplayValue_ReturnValue_Offset), 0);
＋            return returnValue; // missing　
            }
        }
```

If it helps anything, I've put together a [minimal project](https://github.com/CuppaXanax/InkpotUnrealSharpBrk) reproducing the issue.

OlssonDev:
I can't seem to reproduce the issue in the project you provided

CuppaXanax:
Odd - I just tried re-opening after pulling from `HEAD` and I'm still seeing the error during project startup:

<img width="1811" height="691" alt="Image" src="https://github.com/user-attachments/assets/118058eb-ed94-4f6d-916a-6ffaf1ef2ef2" />

I made sure to delete binaries and intermediates; I'm running from VS Code if that matters any.

Any tips on investigating/debugging this myself? I'm happy to make a PR if I can root cause it without too much trouble :)

CuppaXanax:
[...]

PR #549, opened 2025-09-22
Fix generation for cases where property getter & setter differ in their types.

CuppaXanax:
- Property pairing was dependent on accessor discovery order. When the setter was encountered first, the property type came from the setter parameter, and the getter wrapper omitted a return if the return type differed — generating invalid code (empty getter).
- Prefer the getter’s value type (return or out) as the canonical property type and bind both getter and setter exporters to it. Always emit a return in getter wrappers (implicit conversions handle mismatches like FText->string).

Fix for #459


OlssonDev:
Thank you for the fix! :D

PR #550, opened 2025-09-25
Ensure trace channel names starting with digits are prefixed with '_' to prevent invalid C# identifier generation

GameMill:
A fix for invalid C# glue generation resulting from a trace name starting with a digit.

https://discord.com/channels/1153570806417342466/1299809051965657208/threads/1420602750688952361


OlssonDev:
Thank you for the fix! :D

PR #551, opened 2025-09-25
Fix C#-defined structs not using copy marshallers for collections

thomasfn:
Collection members in structs should be marshalled using copy-semantics, as the C# representation of collections are reference types and so may live on past the scope of the uproperty that holds the struct. The glue generator gets this right, however the weaver does not. This fixes the weaver to use copy marshallers for structs.

 Example:
```CSharp
[UStruct]
public struct FMyMap
{
    [UProperty]
    public IDictionary<int, int> Map;
}

...

// In some UClass
[UProperty]
public FMyMap MyMap { get; set; }

...

MyMap = new() { Map = new Dictionary<int, int>() { { 1, 2 } } };
MyMap = new() { Map = new Dictionary<int, int>() { { 2, 3 } } };

// MyMap should contain only [2, 3] but instead contains both [1, 2] and [2, 3]
```

Also, marshalling a map into a uproperty was not clearing the underlying map, resulting in the map being merged instead of overwritten. This fixes that by adding a clear before the copy. The only thing a bit weird here is if you set a property to itself, which will clear it instead of noop-ing, e.g.

```CSharp
[UProperty]
public TMap<int, int> MyMap { get; set; }

...

MyMap.Add(1, 2);
MyMap = MyMap; // Clears it
```

This is kind of a weird thing to do, and I'm not sure if having a setter on a TMap property even makes sense, but maybe we should handle it?

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/MapBase.cs
+++ b/Managed/UnrealSharp/UnrealSharp/MapBase.cs
@@ -425,7 +425,8 @@ public class MapMarshaller<TKey, TValue> where TKey : notnull
     public void ToNative(IntPtr nativeBuffer, int arrayIndex, IDictionary<TKey, TValue> value)
     {
         TMap<TKey, TValue> wrapper = MakeWrapper(nativeBuffer);
-        
+
+        wrapper.Clear();
         foreach (KeyValuePair<TKey, TValue> pair in value)
         {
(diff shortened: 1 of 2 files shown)


OlssonDev:
Awesome! I noticed this too when rewriting the weaver for a source generator! Thank you for the fix :D

PR #554, opened 2025-10-02
Add static load methods for classes and objects

circleapps:
Introduces StaticLoadClass and StaticLoadObject methods to UObject and exposes corresponding interop and native functions. These allow loading classes and objects by asset path, improving dynamic asset management capabilities.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/Extensions/CoreUObject/Object.cs
+++ b/Managed/UnrealSharp/UnrealSharp/Extensions/CoreUObject/Object.cs
@@ -191,28 +191,75 @@ public partial class UObject
     /// <param name="path">Asset path of the class (e.g., "/Game/Path/To/MyClass.MyClass").</param>
     /// <param name="outer">Optional outer object.</param>
     /// <returns>A TSubclassOf wrapping the loaded class.</returns>
+    /// <exception cref="ArgumentException">Thrown if the path is null or empty.</exception>
+    /// <exception cref="InvalidOperationException">Thrown if the class cannot be loaded or the native class cannot be found.</exception>
     public static TSubclassOf<T> StaticLoadClass<T>(string path, UObject? outer = null) where T : UObject
     {
+        if (string.IsNullOrWhiteSpace(path))
+        {
+            throw new ArgumentException("Path cannot be null or empty.", nameof(path));
+        }
+
         IntPtr basePtr = typeof(T).TryGetNativeClass();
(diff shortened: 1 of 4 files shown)


FailMelon:
Why do we need StaticLoadClass/Object exposed instead of using soft references, i can think of maybe editor tooling? Soft refs are the recommended for runtime gameplay code since they integrate with the AssetManager, and allow async loading. StaticLoad* is blocking. If we do include this, it should be clearly marked as unsafe and probably editor-only with guardrails (path validation, type checks)

circleapps:
I added this mainly to reduce redundant configuration through naming conventions.

For example, in the Cropout project, each Menu Widget corresponds to a specific Widget Blueprint Asset. To display a UI from C# code, we need to load the associated asset via a Soft Reference. But this setup feels repetitive and verbose.

My current approach uses the class name of the Menu Widget to dynamically locate and load its corresponding Widget Blueprint Asset. This reduces a lot of manual configuration and aligns with practices seen in frameworks like Ruby on Rails, which also leverage naming conventions to streamline development.

I agree that async loading is a benefit for using soft reference. I am not sure whether we can have both dynamic and async loading,  but I can look into it.

FailMelon:
Naming-by-convention makes sense, but StaticLoad isn’t cook-safe. You can usually get the same benefit with soft refs (e.g. a generated registry or PrimaryAsset IDs), which keeps async/cook safety. With StaticLoad, Assets that are only referenced via StaticLoad are not guaranteed to be included in a packaged build. They’ll be excluded unless they’re also pulled in by something the cooker can see like hard references from other cooked assets, Asset Manager/Primary Asset rules, “AlwaysCook”/Additional Asset Directories, ini-based includes, etc.

FailMelon:
@OlssonDev

FailMelon:
Accidently closed with comment sorry

circleapps:
Thanks for the feedback! I've made further adjustments to enhance error handling and added an asynchronous loading version. The issue of Cooked content not including dynamically loaded assets already exists in UE by design. If you think it's important to explicitly remind users of this, I'm happy to add some additional comments.

OlssonDev:
While I think these are still good bindings for some editor tooling, using it for gameplay is not recommended. I don't like to compare UE development with other workflows such as you might find in web development, it's a different beast and should be treated different. Referencing assets is done by UProperty and it'll be easy on everyone in your team.

PR #562, opened 2025-10-09
Fallback default DotNET on Windows

RPG3D:


OlssonDev:
Why do we need this? Has it been a bug that it can't find it?

RPG3D:
> Why do we need this? Has it been a bug that it can't find it?

on my pc, the PATH settings is "C:\Program Files\dotnet", no postfix \
maybe it is set manually some time ago

PR #564, opened 2025-10-11
Fix build failure

hez2010:
- `_function` has been renamed to `Function`, we should update the reference
- `UE5Rules_DefineConstants` contains an unexpanded text `$(DefineConstant)` which should be removed from `DefineConstantsJoined`, otherwise it will be passed in an unexpanded state to the compiler which will lead the whole `/define` argument to be ignored and result in build error.


OlssonDev:
Thank you for the fix :D

PR #565, opened 2025-10-13
Make delegate trimming and AOT safe

hez2010:
Instead of looking up the wrapper type by `U{name}`, emit an attribute and save the wrapper type into a `Type` property that is marked with `DynamicallyAccessedMemberTypesAttribute`, so that the wrapper type won't be trimmed by the trimmer or the AOT compiler:

```cs
[UMultiDelegate]
delegate void Foo();
```

becomes

```cs
[UMultiDelegate(WrapperType = typeof(UFoo))]
delegate void Foo();

class UFoo : ...
```

then the wrapper type can be obtained by `typeof(Foo).GetCustomAttribute<UDelegate>().WrapperType`.

This is also a performance improvement.

Also replacing some `Marshal.SizeOf(Type)` with `Marshal.SizeOf<T>()` to make it aot-compatible.

The others are some minor fixes to comments/nullable issues.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/Array.cs
+++ b/Managed/UnrealSharp/UnrealSharp/Array.cs
@@ -127,7 +127,7 @@ public class TArray<T> : UnrealArrayBase<T>, IList<T>, IReadOnlyList<T>
         int numElements = Count;
         for (int i = 0; i < numElements; ++i)
         {
-            if (this[i].Equals(item))
+            if (EqualityComparer<T>.Default.Equals(this[i], item))
             {
                 return i;
             }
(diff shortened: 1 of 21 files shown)


OlssonDev:
I'm not a big fan of manually typing the wrappertype in the attribute constructor, I'll think of a way we can do this in another way

hez2010:
I don't think any other alternatives can make such a thing trimming safe completely in C# code, at least before .NET 10. Any attempts that try to look up a type with a non-constant string that cannot be determined at compile time are considered to be trimming-unsafe, and they can not be guaranteed to work at runtime. 
The approach in this PR is also the same approach as CsWinRT, where in CsWinRT, it also needs to associate a type with its projection type: https://github.com/microsoft/CsWinRT/pull/1224.

.NET 10 introduces type mapping APIs so that you can emit some attributes to the assembly that associate a type with another type, and then you can get a dictionary for type lookup at runtime. This is the officially recommended way to do this, and this allows you to do the attribute emitting with a source generator, but this would require at least .NET 10. For details, see https://github.com/dotnet/runtime/issues/113362.

OlssonDev:
I think I have a option where we can do it without, I'll fiddle around a little bit with it :D

Sorry for the delayed response. Been focusing on the source gen

PR #566, opened 2025-10-14
Optimized Allocation Overhead of ConfigureWithUnrealContext on ValueTask

retroandchill:
Originally calling ConfigureWithUnrealContext on ValueTask would use the AsTask method to configure the synchronization context, however by doing so it loses out on one of ValueTask's key benefits, that being the ability of it to complete synchronously without incurring an allocation. So to ensure that stays the case this PR changes the ValueTask overloads to do the following:

- Returns ValueTask instead of Task
- If the task is complete then simply return the parameter back to the caller
- Otherwise create a new value task that wraps the result of the original implementation

By doing this you ensure that your tasks continue on the correct thread (assuming they are started from the game thread, which is almost always going to be the case), while also not allocating on a synchronous completion.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/UnrealSynchronizationContext.cs
+++ b/Managed/UnrealSharp/UnrealSharp/UnrealSynchronizationContext.cs
@@ -80,8 +80,11 @@ namespace UnrealSharp
             });
         }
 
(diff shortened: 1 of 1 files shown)


OlssonDev:
Thank you for the fix! :D

PR #567, opened 2025-10-14
Improved the API of TOptional to resemble that of a fully-fledged optional type

retroandchill:
The new TOptional type does a lot to reduce the reliance on LanguageExt.Core, but lacks a lot of features that we see in other languages. This PR seeks to remedy that, while also allowing for more freedom of use than the previous iteration. Many of these changes are inspired by [Java's Optional class](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Optional.html).

These changes include:
- Calling Value now throws when the optional is not set
- Marked the generic parameter of notnull, which warns about nullable reference types and blocks nullable value types from being used as the type parameter
- Made TOptional's constructor private, forcing the user to go through Some to set a value.
- Added a TryGet method that is typical of C# APIs
- Added transformation and filtering operations

A future extension point for this may be to create an analyzer to warn about accessing Value without checking HasValue first, or suggesting the user opt to use TryGet instead of checking and then accessing, as is seen in other cases by code suggestion engines.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/Optional.cs
+++ b/Managed/UnrealSharp/UnrealSharp/Optional.cs
@@ -1,44 +1,128 @@
 ﻿
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using UnrealSharp.Core;
 using UnrealSharp.Core.Marshallers;
 using UnrealSharp.Interop;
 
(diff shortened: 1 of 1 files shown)


OlssonDev:
Thank you for the fixes! :D

PR #574, opened 2025-10-25
Fixed InvokeNativeFunction to handle out parameters

FailMelon:
This PR fixes an issue with invoking native functions with out parms and Array based parms (They are required to be out parms) as they were not being added to the FFrame stack.


OlssonDev:
Thank you for the PR :D

PR #586, opened 2025-11-11
Update settings viewer to use dynamic names

dahai-li-1993:
bug fix for "open settings" that it did not direct to the correct page.


OlssonDev:
Thank you for the fix:D

PR #588, opened 2025-11-20
Disambiguate delegates with same name but different owners

circleapps:
Delegates with identical names but different owner classes (e.g., UComboBoxString vs UComboBoxKey) are now uniquely identified using their owner's name as a prefix. This prevents export and file overwrite conflicts, and ensures correct attribute generation and file naming for delegates. Utility and exporter logic was updated to support this naming scheme.


OlssonDev:
Thank you for the fix! :D Did merging fixes now when the new source gen branch is in

PR #593, opened 2025-12-11
Include-references needs to be done after the project directories to avoid generation errors

EchoEclipseWolf:
--include-references needs to be done after the project directories you are adding otherwise it throws the error Could not find project or directory `--include-references`. Does this change depend on the dotnet version though? Do we need a version check here or is this safe?

The following error is thrown in its current state
System.Exception: BuildTool process failed with exit code 1:
Could not find project or directory `--include-references`.
Could not find project or directory `false`.

Which results in people reporting 
Attempt 3/10: error while adding projects to sln: BuildTool process failed with exit code 1:
Could not find project or directory `--include-references`.
Could not find project or directory `false`.

dotnet --info
.NET SDK:
 Version:           9.0.306
 Commit:            cc9947ca66


OlssonDev (approved):

OlssonDev:
Thank you for the fix :D

PR #597, opened 2025-12-14
Improve diagnostic reporting in UnrealTypeDiscoveryGenerator & Fixes Could not load file or assembly 'System.Text.Json...

GameMill:
Enhance error reporting by including exception message.


GameMill:
Because VS and Rider both limit error messages to a single line, I've chosen to throw a second error containing the exception message so that errors not specifically related to the source generator are also included.

For example, on a new project I'm getting
"Could not load file or assembly 'System.Text.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified." from the exception message.

GameMill:
I've added CopyLocalLockFileAssemblies to UnrealSharp.GlueGenerator.csproj to ensure required files are copied during the build. 

This resolves the issue:
Could not load file or assembly 'System.Text.Json, Version=9.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' or one of its dependencies. The system cannot find the file specified.

GameMill:
The issue has returned, so going to close this PR and do some more digging

PR #606, opened 2025-12-19
Fix editor crash when calling a newly-created RPC (NetServer) for the first time

meglinge:
## Problem

- In Editor, a newly-created pure RPC (e.g. `RunOnServer|Reliable` without `BlueprintEvent`) can crash on its very first call.
- Adding `BlueprintEvent` once “fixes it forever”, suggesting the initial function resolution/caching path is wrong for pure RPCs.

## Root Cause

- In Editor, managed type resolution frequently points to `SkeletonGeneratedClass`.
- UnrealSharp’s generated glue caches `UFunction*` during the managed type’s static constructor via `GetNativeFunctionFromClassAndName(Class, Name)`.
- For Net/RPC functions, a `UFunction*` coming from the skeleton can lack valid runtime replication metadata (e.g. `RPCId` / callspace-related data), causing UE’s RPC dispatch (`GetFunctionCallspace` / `CallRemoteFunction`) to crash on first use.

## Fix Summary

- Ensure RPC calls use the `UFunction` resolved from the runtime class / generated class, not the skeleton.

## Changes

- `Source/UnrealSharpCore/Private/Export/UClassExporter.cpp`
  - Editor-only: if the initially found function is `FUNC_Net`, prefer returning the same-named `UFunction` from the corresponding `GeneratedClass` (covers `UCSSkeletonClass` and regular `UBlueprintGeneratedClass`).
  - Preserve the previous behavior for non-Net functions (still prefer the provided class, which is important for editor compilation where many functions only exist on the skeleton).
  - Keep an editor-only fallback: if the function is not found on the provided class, try the generated class.

- `Source/UnrealSharpCore/Private/Export/UObjectExporter.cpp`
  - Editor-only: before invoking a Net function, re-resolve `UFunction` by name on the object’s runtime class (`NativeObject->FindFunction`).
  - If the runtime class is a skeleton (`UCSSkeletonClass`), additionally jump to its `GeneratedClass` and resolve the same-named function there.

## Testing

- Compile:
  - `Build.bat CapProjectEditor Win64 Development -Project="E:\UE\CapProject\Client\CapProject.uproject"`
  - Note: linking `UnrealEditor-UnrealSharpCore.dll` / `UnrealEditor-UnrealSharpCompiler.dll` will fail if `UnrealEditor.exe` is running and holding the plugin DLLs; close the editor before building.
- Manual repro (expected):
  - Create a brand new `[UFunction(FunctionFlags.RunOnServer | FunctionFlags.Reliable)]` and call it immediately in `BeginPlay();` should no longer crash on first call in Editor.

## Impact / Risk

- Editor-only behavior (`WITH_EDITOR` guards); packaged/runtime builds should be unaffected.
- Non-Net function lookup behavior is preserved to avoid breaking skeleton-only editor compilation paths.


meglinge:
ba69ea72b807f8d93b78ba19898f144f67e89cd2
c7032c3c3ea29f3dec9ba3e5843c34945a072315
## Problem

When a brand-new RPC (e.g. `RunOnServer | Reliable` without `BlueprintEvent`) is called for the first time after opening the editor, PIE becomes extremely slow for the whole session. If you stop PIE and run again, it becomes smooth. After restarting the editor, the first PIE is slow again.

## Root Cause (high-level)

UnrealSharp generated C# code was resolving and caching the `UFunction*` too early (during type static initialization) using a class pointer that often points to the skeleton class in the editor.

For newly created “pure RPC” functions, the skeleton path does not have complete/valid runtime networking/replication metadata. This causes UE to do expensive “fix-up / cache build / resolution” work the first time, and in some cases it can effectively poison the whole first PIE session (lots of heavy work or repeated resolution/logging), while the second PIE is fast because the engine has cached everything.

## Why the fix removes the “first PIE hitch”

- For network functions (RPCs) we stopped caching the invocation `UFunction*` during static initialization (when we’re most likely operating on a skeleton class).
- Instead, the wrapper now lazily resolves the `UFunction*` from the runtime instance on first call (which resolves against the generated/runtime class), then caches it so subsequent calls are cheap.
- On the native side, any editor-only “RPC function re-resolution” is restricted to cases where the cached function is actually owned by UnrealSharp’s skeleton class, avoiding unnecessary lookups for normal inheritance and preventing per-call overhead that could keep the entire PIE session sluggish.

OlssonDev:
This might already be fixed on main. I looked into it, and it seems like the changes here are unnecessary. 

The RPCs are events, but not BlueprintEvents.

I'll close, and we can re-open if it's not resolved. But it now works on main

PR #608, opened 2025-12-19
Fix garbled Chinese text when passing C# strings into Unreal

meglinge:
## Summary

Fixes Unicode data loss when C# `string` values are passed into Unreal via UnrealSharp. The previous interop path relied on default runtime marshalling to `char*` (ANSI/ACP), which replaced non-ASCII characters with `?`, causing garbled output in UMG, `PrintString`, and logs.

This change pins the managed UTF-16 buffer and passes a `(char* ptr, int length)` “string view” into native code, where it is converted to `TCHAR` safely.

## Key Changes

- Added a new native exporter API to marshal `FString` from a UTF-16 view:
  `UFStringExporter::MarshalToNativeStringView(FString*, const UTF16CHAR*, int32 Length)`
- Updated `StringMarshaller` to use the UTF-16 view path instead of passing `string` directly to native.
- Updated `FText(string)` to call `FromStringView(char*, len)` (pinned UTF-16) rather than `FromString(string)` to avoid ANSI marshalling.

## How To Test

- Rebuild UnrealSharp managed assemblies and the plugin.
- In C#, call:
  `UObject.PrintString("中文测试：你好世界")`
- Set a UMG `TextBlock` text from a C# string containing Chinese characters.
- Log a Chinese string via UnrealSharp logging APIs.
- Verify the text displays correctly (no `???` or mojibake).

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp.Core/Interop/FStringExporter.cs
+++ b/Managed/UnrealSharp/UnrealSharp.Core/Interop/FStringExporter.cs
@@ -6,4 +6,5 @@ namespace UnrealSharp.Core.Interop;
 public unsafe partial class FStringExporter
 {
     public static delegate* unmanaged<UnmanagedArray*, string, void> MarshalToNativeString;
+    public static delegate* unmanaged<UnmanagedArray*, char*, int, void> MarshalToNativeStringView;
 }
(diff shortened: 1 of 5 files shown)


OlssonDev:
Thank you for the fix! :)

PR #612, opened 2025-12-21
Json newtonsoft conversion

FailMelon:
This PR converts System.Text.Json in the glue generator as in .netstandard it has to many dependencies to deal with, this also fixes an issue with user projects showing up that the analyzer is missing the dll.

Changes with this PR also include going from JObject to TextWriter as this reduces memory usage and cpu usage by removing the need to build an in-memory json dom and creating allocations with JObject JValue etc.


OlssonDev:
Thank you for the fix :D

PR #613, opened 2025-12-22
Disable rollforward on runtime processes

FailMelon:
small PR that fixes an issue where USharpBuildTools will run on a different version of .net runtime, i've tried to disable rollForward in the runtimeconfig.json as well in the csproj however it didn't seem to fix so ended up with setting it as an environment variable instead.


OlssonDev:
Thank you for the fix :D

PR #618, opened 2025-12-23
Update DotNetUtilities.cs

dahai-li-1993:
## Problem
unrealsharp was forcing to use .net 9.0.0 due to the fact that rollforward was disabled.


FailMelon:
Sorry didn't see this already bumped it.

PR #621, opened 2025-12-26
Fixes global namespace

FailMelon:
Currently there is an issue with global namespaces where it'll return `<global Namespace>` as the Namespace, this PR makes namespaces empty if global and then removes namespace entirely from the FieldName.FullName if the namespace is empty.


OlssonDev:
Nice! Thank you for the fix :D

PR #623, opened 2025-12-28
Fixes net call absorbing local invoke

### LINKED ISSUE #622: I encountered a problem while using RPC communication.
When the client calls an RPC function marked with "Server" and this function internally calls a Multicast broadcast, only the client itself receives the broadcast. Initially, I thought I might have made a mistake somewhere, but after checking for a long time, I couldn't find anything. 
I tried rewriting it in C++ and found that it worked normally—both the server and the client could receive the broadcast. The related code is as follows:

C++ side
```C++
UCLASS(Blueprintable, BlueprintType)
class AHomelandGameState : public AGameState
{
	GENERATED_BODY()
public:
	UFUNCTION(BlueprintCallable, Server, Reliable, Category = "HomelandGameState")
	void Server_SetAdventureRoomReadyState(const bool bIsReady);

	UFUNCTION(BlueprintCallable, Reliable, NetMulticast)
	void Multicast_AdventureRoomReadyStateChanged(const bool bIsReady);

	void Server_SetAdventureRoomReadyState_Implementation(const bool bIsReady)
	{
		Multicast_AdventureRoomReadyStateChanged(bIsReady);
	}
	
	void Multicast_AdventureRoomReadyStateChanged_Implementation(const bool bIsReady)
	{
		UE_LOG(LogTemp, Log, TEXT("here in C++, auth:%d"), GetWorld()->GetNetMode());
	}
};
```
The output is 
```
LogTemp: here in C++, auth:2
LogTemp:here in C++, auth:3
```

C# side.  I commented out all the content of AHomelandGameState on the C++ side and rewrote it in C#.
```C#
[UClass]
public partial class ABusyHomelandGameState : AHomelandGameState
{
    [UFunction(FunctionFlags.RunOnServer|FunctionFlags.Reliable)]
	public partial void Server_SetAdventureRoomReadyState(bool bIsReady);

    [UFunction(FunctionFlags.Multicast|FunctionFlags.Reliable)]

	public partial void Multicast_AdventureRoomReadyStateChanged(bool bIsReady);

	public partial void Server_SetAdventureRoomReadyState_Implementation(bool bIsReady)
	{
		Multicast_AdventureRoomReadyStateChanged(bIsReady);
	}

	public partial void Multicast_AdventureRoomReadyStateChanged_Implementation(bool bIsReady)
	{
        LogAdventureRoom.Log($"here in C#, auth:{World.GetNetMode()}");
	}
}
```

The output is 
```
LogAdventureRoom: Display: here in C#, auth:Client
```

FailMelon:
This PR fixes an issue where a net callspace can be a flag of both local and remote, right now if anything is flagged with remote it immediately returns. This causes issues with multicast as they are called on the server as well thus the callspace being assigned to 3 aka local and remote. We now check to see if we don't have the local Callspace flag this is currently how ScriptCore also handles it.

Fixes #622

### DIFF
--- a/Source/UnrealSharpCore/Private/Export/UObjectExporter.cpp
+++ b/Source/UnrealSharpCore/Private/Export/UObjectExporter.cpp
@@ -75,6 +75,10 @@ void UUObjectExporter::InvokeNativeNetFunction(UObject* NativeObject, UFunction*
 	if (FunctionCallspace & FunctionCallspace::Remote)
 	{
 		NativeObject->CallRemoteFunction(NativeFunction, Params, nullptr, nullptr);
+	}
+
+	if ((FunctionCallspace & FunctionCallspace::Local) == 0)
+	{
 		return;
 	}
 


OlssonDev:
Thank you for the fix :D

PR #624, opened 2025-12-29
Fixes returns in glue generator

FailMelon:
This PR fixes an issue in glue gen when exporting from native in return where it would just apply everything to `returnValue =` this fixes that by passing in as an assignment operator instead.


OlssonDev:
Thank you for the fix :D

PR #629, opened 2026-01-10
Fixes order of key value being reversed with new container setup

FailMelon:
This PR fixes an issue where the key and value on map containers are reversed.


OlssonDev:
Thank you for the fix :D

PR #630, opened 2026-01-14
Fix crash when failed to find type definition

GameMill:
Add IsValid check to avoid crashes when ManagedTypeDefinition is null.
Remove UE type-name prefix that caused ManagedTypeDefinition to be null when using UObject.AsInterface<T>().

### DIFF
--- a/Source/UnrealSharpCore/Private/Export/UCoreUObjectExporter.cpp
+++ b/Source/UnrealSharpCore/Private/Export/UCoreUObjectExporter.cpp
@@ -19,7 +19,8 @@ UField* UUCoreUObjectExporter::GetType(const char* InAssemblyName, const char* I
 	FCSFieldName FieldName(InTypeName, InNamespace);
 
 	TSharedPtr<FCSManagedTypeDefinition> ManagedTypeDefinition = Assembly->FindOrAddManagedTypeDefinition(FieldName);
-	if (!ManagedTypeDefinition.IsValid()) {
+	if (!ManagedTypeDefinition.IsValid()) 
+	{
 		UE_LOGFMT(LogUnrealSharp, Warning, "Failed to find type definition: {0}.{1} in assembly {2}", InNamespace, InTypeName,
 				  InAssemblyName);
 		return nullptr;


OlssonDev:
Can you show the code that causes this?

GameMill:
I can run a BoxTrace and retrieve the hit actor, but I can’t cast it directly because the returned AActor doesn’t allow a normal cast and must use the native interface system. That’s why I call targetActor->AsInterface<IPickupable>() to access the interface.

```
 [UInterface]
 public interface IPickupable : IInterface
 {
     [UFunction]
     public UPrimitiveComponent GetPrimitiveComponent();
 }
 ```
 The code on my character
```
 [UFunction]
 public void Grab(FInputActionValue value)
 {
     if (BaseCharacter is null) return;
     UCameraComponent camera = BaseCharacter.GetActiveCameraComponent();
     FVector endLocation = camera.WorldLocation + (camera.ForwardVector * 250f);
     AActor? targetActor = null;

     if (SystemLibrary.MultiBoxTraceByChannel(camera.WorldLocation, endLocation, new UnrealSharp.CoreUObject.FVector(10, 10, 10), camera.RelativeRotation, ETraceTypeQuery.TraceTypeQuery1, false, default, EDrawDebugTrace.ForOneFrame, out IList<FHitResult> outHits, true))
     {
         if(outHits[0].Actor == null && outHits.Count > 1)
         {
             outHits = outHits.Skip(1).ToList(); // the first is sometimes a non existent actor? U# issue?
         }
         targetActor = outHits[0].Actor;
     } else
     {
         targetActor = HeldActor;
     }

     if (targetActor == null)
     {
         DropHeldActor();
         return;
     }

     if (HeldActor != null & HeldActor != targetActor)
     {
         DropHeldActor();
         return;
     }

     if (targetActor.ActorLocation.IsNear(camera.WorldLocation, 500))
     {
         HeldActor = targetActor; // targetActor sometimes becomes null after assigning HeldActor? GC?
         if(targetActor == null) return;
         var pickupableActor = targetActor.AsInterface<IPickupable>(); // This is where the errors happens
         UPrimitiveComponent? primitiveComponent = null;
         if (pickupableActor != null)
         {
             primitiveComponent = pickupableActor.GetPrimitiveComponent();
         } else
         {
             primitiveComponent = targetActor?.RootComponent as UPrimitiveComponent;
         }
         if (primitiveComponent != null)
         {
             targetActor.CustomTimeDilation = BaseCharacter.CustomTimeDilation;
             primitiveComponent.EnableGravity = false;
             SetVelocity(endLocation, primitiveComponent, outHits);
             SetRotation(camera, targetActor);
             primitiveComponent.SetAllPhysicsAngularVelocityInRadians(FVector.Zero, false);
         }
     }
 }
 ```


image

image
Type names contains the UE prefix
image
Defined managed types are stored without the UE prefix
image
leading to nullptr
image

and because of the nullptr it crashes the editor on the next line
image

OlssonDev:
Right. I need to introduce partial interface for that attribute. 

Completely forgot about that when doing the source gen

OlssonDev:
Closing this now as I instead added partial interfaces requirement that generates the attribute

PR #633, opened 2026-01-23
Fixed Crash when using Localization Dashboard

lesij-dev:


OlssonDev:
Awesome! Better to make it return false in `ShouldCreateSubsystem` if running commandlet

PR #634, opened 2026-01-24
fixed crash when using Localization Dashboard

lesij-dev:
rel #633


OlssonDev:
Thank you for the fix!

PR #636, opened 2026-01-26
Add buffer to reserved words to avoid collisions with local variable names

Kethku:
When using UnrealSharp with Hazelight's Angelscript fork of unreal, the glue generation fails due to a function they added which has an argument named InBuffer. This variable gets renamed to buffer which collides with a local variable defined later in the same generated glue function.

This fixes that by making buffer a reserved variable name.


Kethku:
I'm unsure if this is the approach you'd prefer because its not really a keyword in the strict sense. Maybe a more correct solution would be to use a more obviously non colliding variable name for the intermediate variable in the generated glue function. If that would be preferable I can make that change instead.

OlssonDev:
Thank you for the fix!

PR #637, opened 2026-01-26
[WIP] added editor settings to hide warnings in build tool

dahai-li-1993:
image

becomes...

image

I think it is more practical because errors are mantory to be fixed to initialize the edtor. Meanwhile there are too often having null reference, which is already exempted by calling "IsValid()"

### DIFF
--- a/Source/UnrealSharpProcHelper/Private/CSProcUtilities.cpp
+++ b/Source/UnrealSharpProcHelper/Private/CSProcUtilities.cpp
@@ -1,6 +1,7 @@
 ﻿#include "CSProcUtilities.h"
 
 #include "CSCommonUnrealSharpSettings.h"
+#include "CSUnrealSharpProcHelperSettings.h"
 #include "CSUnrealSharpSettingsUtilities.h"
 #include "UnrealSharpProcHelper.h"
 #include "Misc/App.h"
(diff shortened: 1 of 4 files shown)


FailMelon:
I think we could do this instead with passing -clp:ErrorsOnly, or -v:minimal maybe instead of stripping lines?

dahai-li-1993:
This PR is ready to merge. Please squash and merge when convenient.

Source/UnrealSharpProcHelper/Private/CSProcUtilities.cpp
```diff
@@ -140,6 +141,20 @@ bool UCSProcUtilities::BuildUserSolution()
 {
 	TMap<FString, FString> Arguments;
 	Arguments.Add("OutputPath", GetUserAssemblyDirectory());
+	
+	bool bShowBuildWarnings = true;
+	GConfig->GetBool(
```
OlssonDev: We should probably make a developer settings for proc helper instead.

FailMelon: i think realisticly it should probably end up as a editor setting and the initial build of the project should probably also happen in the editor module, but its a lot of work outside the scope of this pr to move.

PR #641, opened 2026-02-01
Improve world context stability

FailMelon:
This PR helps improve the world context stability by using the World of the UObject instead of the UObject. Sometimes there are edge cases where the UObject will get GC'd making the WorldContext becomes invalid. By using the World instead i hope to improve stability as it seems less likely that the world will be GC'd. :smile:


OlssonDev:
Thank you for the fix :D

PR #643, opened 2026-02-01
fixed capital error in gitignore

lesij-dev:


OlssonDev:
Thank you for the fix! :D

PR #644, opened 2026-02-02
Convert ECSWorldType to enum class to avoid 'redefinition'-errors.

hallgaeuer:
On current main with UE 5.7.2 I experienced compile errors caused by the ECSWorldType enum.

```
CSWorldExtensions.h(30,2): Error C2365 : "None": Erneute Definition; vorherige Definition war "Enumerator".
	None = EWorldType::None,
	^
```
(Sorry for the german, in english it would be `error C2365: 'None': redefinition; previous definition was 'enumerator'`)

I could reproduce this on an new and clean 5.7.2 project.
Seems like some people also reported the same issue on discord.

This PR fixes this issue by converting ECSWorldType to an enum class instead. 

The existing static_casts seem to work fine in my testing. 
I replaced the old static_assert with static_asserts checking each enum value. Not sure how we could catch epic adding new values to their enum though.

### DIFF
--- a/Source/UnrealSharpCore/Public/Extensions/Libraries/CSWorldExtensions.h
+++ b/Source/UnrealSharpCore/Public/Extensions/Libraries/CSWorldExtensions.h
@@ -51,15 +51,6 @@ enum class ECSWorldType : uint8
 	Inactive = EWorldType::Inactive,
 };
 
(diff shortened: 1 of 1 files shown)


hallgaeuer:
Actually, now that I think about it, those static_asserts I added are probably useless.. :) The values are assigned directly before, so they would always be true anyway. I will remove them..

OlssonDev:
Thank you for the fix!

PR #645, opened 2026-02-02
Force `dotnet sln new` to use `sln` format instead of `slnx` for now

hallgaeuer:
I have both dotnet 9 and 10 SDKs installed. This causes the `sln new` command used by `Managed/UnrealSharpPrograms/UnrealSharpBuildTool/Actions/GenerateSolution.cs` to generate an `slnx` file instead of `sln` (new default in dotnet 10). 
Some other place still seems to generate an `sln` file. This then causes "multiple project files found" errors.

I experienced these errors when creating a completely new U# project on UE 5.7.2.

This PR forces the format to `sln` for now until you decide to make make `slnx` the new default :)


FailMelon:
I believe the plan is to switch over to slnx it just hasn't been done yet, but for now we can merge this until we have transitioned over so that the project is in a workable oob state. Thanks for the PR!

Issue #642, opened 2026-02-01
Generic type parameter mismatch: CreateWidget_WithParent<T>(TSubclassOf<T>, UUserWidget) becomes CreateWidget_WithParent<T>(TSubclassOf<T>, T)


dahai-li-1993:
```
public static T CreateWidget<T>(TSubclassOf<T> widgetClass, UUserWidget? owningWidget) where T : UUserWidget
{
    return UCSUserWidgetExtensions.CreateWidget_WithWidget(widgetClass, owningWidget);
}
```

The underlying generated method signature is:
`
CreateWidget_WithWidget<T>(TSubclassOf<T>, T)
`

The overload should allow creating a widget of type T while passing any UUserWidget as the parent/owner reference.

This might be a case for all DOT methods

PR #649, opened 2026-02-06
Fix generic substitution leakage to non-DOT/DOP parameters

dahai-li-1993:
This is a pre-fix for https://github.com/UnrealSharp/UnrealSharp/issues/642

Before this change, `IsGenericType()` effectively said:
- function has DOT/DOP metadata, and
- this property’s managed type equals DOT’s managed type

then treat it as generic (DOT).
That was too broad. In CreateWidget_WithWidget<T>(TSubclassOf<T> widgetClass, UUserWidget owningWidget), owningWidget is also UUserWidget, so it got incorrectly rewritten to T (DOT) even though it is not part of the DOT/DOP pair.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,5 +82,5 @@ obj/
 
 *.props
 !Directory.Packages.props
-!Directory.Build.Props
-!UnrealSharp.Shared.Props
+!Directory.Build.props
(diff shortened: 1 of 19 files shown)


OlssonDev:
Thank you for the fix! :D

PR #652, opened 2026-02-07
fixed mixed solution merge logic

dahai-li-1993:
The old implementation was outdated and had several flaws:

  1. Brittle path rewriting — split on ", " which breaks if project names/paths contain commas, then blindly
  prepended "Script\" assuming all .csproj files live there
  2. No build configuration merging — C# projects were inserted as project entries but had zero
  ProjectConfigurationPlatforms entries, so Visual Studio would show them as unbuildable
  3. No NestedProjects merging — solution folder hierarchy from the managed solution was lost entirely
  4. No error handling — LastEndProjectIdx initialized to 0, so if no "EndProject" was found, it would silently
  corrupt the file header
  5. Only captured flat lines — only grabbed lines starting with Project("{ or EndProject, missing any
  ProjectSection content between them

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,5 +82,5 @@ obj/
 
 *.props
 !Directory.Packages.props
-!Directory.Build.Props
-!UnrealSharp.Shared.Props
+!Directory.Build.props
+!UnrealSharp.Shared.props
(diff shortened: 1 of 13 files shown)


OlssonDev:
Thank you for the fix! :D

PR #654, opened 2026-02-10
Fix unstable C# hot reload state causing missing in-project types

Extreme11111:
## Summary
- Skip obj/bin path changes in incremental callbacks to avoid watcher noise polluting compilation state.
- Normalize project document and syntax-tree file paths before pruning generated trees.
- Force-refresh project state before source generation during hot reload for stability.

## Problem
On some Windows setups, trivial script saves could trigger hot reload failures with many CS0246 errors for in-project types (for example UVRSharpGrabbable) and produce tiny/broken output assemblies.

## Impact
Hot reload becomes deterministic and avoids partial-compilation states caused by stale incremental state and path normalization mismatches.


OlssonDev:
I can't seem to be able to hot reload with this PR. I do changes but none of them applies once reloaded.

image

image

I tried to change the name of this component and was only fixed once restarted.

Extreme11111:
Thanks, confirmed. My previous change forced ProcessProject() on every hot reload, which reset incremental state and made changes not apply until restart. I’ve updated it to only rebuild state when it is invalid, so normal hot reload remains incremental.

OlssonDev:
Thank you for the PR :D

PR #659, opened 2026-02-28
Managed Glue generation for Struct constructors

FailMelon:
This PR implements automatic managed glue for Struct constructors based on all the possible managed properties found.


OlssonDev:
Thank you for the PR :D Awesome feature!

PR #661, opened 2026-03-01
Add Field Notify Support

FailMelon:
**Requires PR #659**

This PR adds support for Unreal’s FieldNotify system within UnrealSharp, enabling property change notifications to bound UI elements (UMG) or systems.

The following property types are currently supported as FieldNotify properties:
- Simple
- Arrays
- Maps

These property types properly generate and broadcast field notifications when their values change.


OlssonDev:
Thank you for the PR :D

PR #664, opened 2026-03-02
Ability to change Script directory

CENSOR37:
This is only a draft where I think it's being hardcoded, and might help implement them.
If I miss anything, please let me know


CENSOR37:
I have tested with UnrealSharp-Cropout manage to get it running on the editor

OlssonDev:
Nice feature :D Added a general config file for more stuff like this in the future :)

PR #667, opened 2026-03-05
1. fix compile error on MacOS

Unreal-3D:


OlssonDev:
Is the issue the inline?

Unreal-3D:
> Is the issue the inline?

yes

OlssonDev:
Just remove it and implement it in the c++ file

OlssonDev:
Thank you for the fix! :D

PR #668, opened 2026-03-12
Added Interface Blueprint functionallity

lesij-dev:
Please check the Generated code if it can be cleaned up.

Interfaces automatically implement a method to check if an UObject implements that interface in blueprints. Can be called by using the static interface method called Implements.
Each interface function can also call the blueprint implementation directly on the UObject by using the generated static Interface.Execute function.
Why this change?
Example:
You implement an Interface in U#.
A Blueprint implements your interface.
If you want to call your method from the Interface in U# the cast will fail.
This change will allow to call those methods directly on the UObject


OlssonDev:
This is already a feature in U#. You can do `IMyInterface interfaceWrapper = myObject.AsInterface<IMyInterface>` and call BP only implemented interfaces

lesij-dev:
This doesnt work for BP only interfaces (tried it already). Those can not be casted into Interfaces, since they are not actually implemented in code (atleast afaik). Epic also sais you need to use the Interface::Execute_Function methods
[described here](https://dev.epicgames.com/documentation/en-us/unreal-engine/interfaces-in-unreal-engine#call-a-blueprint-event-from-c)

You can try the setup as following:
create interface in u#
implement in bp only -> bp parent is Actor
assign a class reference in an other Object. 
Spawn Actor by this reference
cast into the interface (using the AsInterface method)
result should be an empty cast (null)

lesij-dev:
here is a test
image

image

OlssonDev:
I think I want to look into this, I'd rather much add support to the AsInterface to support these cases.

OlssonDev:
I added support for user created interfaces in AsInterface now. Merging! :D

PR #669, opened 2026-03-15
Fixes a crash when logging a failure to generate a project.

Y0SH1M4S73R:
`UE_LOGFMT` expects its parameters to be indexed or named using curly braces. The use of this macro in `UnrealSharpEditorModule::AddNewProject` used `%s` to specify parameters.


OlssonDev:
Nice catch! :D

PR #671, opened 2026-03-17
Exclude classes with 'SKEL_' prefix from FCSCompilerContext::TryFakeNativeClass

hallgaeuer:
It is my understanding that In `FCSCompilerContext::TryFakeNativeClass` the class flag `CLASS_NATIVE` is added to certain subclasses, so that they show up in certain contexts like the state tree task selection.

Internally unreal generates 'SKEL_' classes in addition to the "regular" classes, which also got the `CLASS_NATIVE` flag in `TryFakeNativeClass`. 
This lead to "duplicates" in the state tree task selection (and possibly other places), because both the SKEL_ class and the "regular" class was shown.

This PR adds a check in `TryFakeNativeClass` to exclude class names starting with "SKEL_". Unreal does this check the same way in some other places internally.

**Before PR:**
grafik

grafik

grafik

**After PR:**
grafik

I also quickly verified that C# BT tasks still show up in the BT task selection. Not sure if they also were "duplicated" before, since I don't use BTs.


OlssonDev:
Thank you for the fix! :D

Issue #227, opened 2024-10-05
[BUG] Framework-provided assemblies not being loaded


ezesil:
To reproduce this issue with in a fresh unreal engine 5 project:

- Open the C# solution
- Add an ASP.NET Core API project
- In the entrypoint/main project: make a new class:

```
[UClass]
public class UGameSubsystem : UCSGameInstanceSubsystem
{
    protected override void Initialize() // This will trigger as soon as the game is run
    {
        base.Initialize();
        APIProject.Program.Main([]);
    }
}
```

- Compile the project in Unreal Engine 5
- Run
- The game should crash and throw an error:

```
Fatal error detected: "System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.AspNetCore, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'. El sistema no puede encontrar el archivo especificado.
File name: 'Microsoft.AspNetCore, Version=8.0.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60'
```

Referencing the DLL from the project works, but it will trigger the same error for each assembly that comes from Framework-Provided assemblies, even if you add the project reference like so:
`<FrameworkReference Include="Microsoft.AspNetCore.App" Version="8.0.0" />`
The error still happens even if they are all in Binaries/Managed.

As a workaround, i extracted all DLLs from C:\<your user>\.nuget\packages\microsoft.aspnetcore.app.runtime.win-x64\microsoft.aspnetcore.app.runtime.win-x64.8.0.6.nupkg file by duplicating it, changing the extension to .zip, and copying all DLLs to the entrypoint project in a new folder, then loading them manually using Assembly.LoadFrom(path) before the API's main call, like so:

LoadAssemblies(); // This loads all DLL dependencies
APIProject.Program.Main([]);

After the call, it will enter Main() method and host the API properly.

I suspect that, given the assemblies are provided by the framework as a whole package, the DLLs are not being found in the same place as the other dependencies are, therefore not found and not loaded.

Klip29mat2:
This looks like an assembly resolution issue rather than missing DLLs.

Framework assemblies like Microsoft.AspNetCore are not copied locally and are expected to be resolved via the .NET runtime.

A possible fix could be:
- Extending the AssemblyLoadContext resolver
- Adding fallback paths (Binaries/Managed, DOTNET shared runtime)
- Or using DependencyContext to resolve framework assemblies properly

[...]

PR #673, opened 2026-04-05
Fix PluginLoadContext Load method

Klip29mat2:
Quick fix for the PluginLoadContext.Load method to handle missing assemblies and fallback to the default load context. Should prevent crashes when loading plugins with dependencies.


OlssonDev:
When does this happen?

Klip29mat2:
it's fix PluginLoadContext, it's for the issue #227

OlssonDev:
Thank you for the fix! :D

PR #677, opened 2026-04-15
try to fix problem: exit game but process remains running in the back…

OriginalSinCat:
The pr is try to fix exit game process problem

Operating System: Windows 10
Unreal Engine Version: 5.7.2
Issue Description:
I have a simple empty project with only an exit button(new project with nothing else). After packaging the game, clicking the exit button does not close the game; the process remains running in the background.

Is this caused by the #if WITH_EDITOR directive around line 163 in CSManager.cpp? 
I tried removing #if WITH_EDITOR and repackaging the game, and the game process now exits normally. 

However, I'm not sure if this modification is correct.

### DIFF
--- a/Source/UnrealSharpCore/Private/CSManager.cpp
+++ b/Source/UnrealSharpCore/Private/CSManager.cpp
@@ -157,11 +157,9 @@ void UCSManager::Initialize()
 		return;
 	}
 
-#if WITH_EDITOR
 	// Remove this listener when the engine is shutting down.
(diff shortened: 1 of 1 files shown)


CanGoktas-Nar:
I'd like to add that, if i don't do this modification, i get a crash in packaged game just like comment says on lines 161 and 162 of `CSManager.cpp` caused by `UObjectArray.cpp` line 369 -> calling a pure virtual function

OlssonDev:
Huh, that's odd! Thank you for the fix! :D

PR #681, opened 2026-05-09
Add support for init only properties

hez2010:
To make things like this work:

```cs
public partial int InitOnlyProp { get; init; }
```


OlssonDev:
That's awesome! Thank you for the PR :D

PR #682, opened 2026-05-13
Fix array indexof comparing null objects

CanGoktas-Nar:
I had a list like `IList<AActor?>(same thing happens without nullable)` and sometimes when i tried to remove a element, `UnrealSharp.sln/Array.cs` line 152 was throwing fatal error: `Object reference not set to an instance of an object` because some element is `null`. This pr aims to fix it. 

Im quite rusty on C# so i don't know if this the correct fix for all cases but fixed the problem i explained.

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp/Array.cs
+++ b/Managed/UnrealSharp/UnrealSharp/Array.cs
@@ -149,7 +149,7 @@ public class TArray<T> : UnrealArrayBase<T>, IList<T>, IReadOnlyList<T>
         int numElements = Count;
         for (int i = 0; i < numElements; ++i)
         {
-            if (this[i].Equals(item))
+            if (this[i] is { } currentItem && currentItem.Equals(item))
             {
                 return i;
             }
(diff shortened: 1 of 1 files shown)


Managed/UnrealSharp/UnrealSharp/Array.cs
```diff
@@ -149,7 +149,7 @@ public int IndexOf(T item)
         int numElements = Count;
         for (int i = 0; i < numElements; ++i)
         {
-            if (this[i].Equals(item))
+            if (this[i] is { } currentItem && currentItem.Equals(item))
```
hez2010: This should use `EqualityComparer<T>.Default.Equals(this[i], item)` instead. 

You might want to check if other collection types like Map, Set etc. have similar issues.

hez2010 (changes requested):

Managed/UnrealSharp/UnrealSharp/Array.cs
CanGoktas-Nar: I didn't encounter same issues with Map and Set. I updated the code to use EqualityComparer. Thanks.

OlssonDev:
Thank you for the fix!

PR #697, opened 2026-06-17
Rename Bind_AsyncExporter to Bind_Async

CanGoktas-Nar:
This pr fixes crash with error `Failed to find bound function RunOnThread in Bind_Async` due to naming issue.


OlssonDev:
Thank you for the fix!

PR #702, opened 2026-06-24
Fix extension methods with DetermineOutputType parameter

Kethku:
I found that extension methods with DetermineOutputType parameter would generate invalid code due to extra commas in the function argument list.

I *think* this is the right fix and in my testing it seems to address the issue but I'm not 100% positive. Hopefully this works.


OlssonDev:
Thank you for the fix :D

PR #703, opened 2026-07-09
Fix static async ufunction crash because of DefaultToSelf metadata

CanGoktas-Nar:
`UnrealAsyncFunction.cs` line 356 adds `DefaultToSelf` metadata to async functions with `Target` value.
This causes crash in `UK2Node_BaseAsyncTask.cpp` line 369 with `static async ufunctions` because there is no `Target pin` .

 Crash happens after connecting input pin and compiling the bp. This pr aims to fix it.


OlssonDev:
Thank you for the fix!

PR #709, opened 2026-07-21
Fix component only compile being skipped during hot reload

### LINKED ISSUE #706: Exposed component variables doesn't reflect in editor until restart
I have a `UActorComponent` that exposes some variables. These variables are not reflected in editor until i restart it.

Steps to reproduce:
1- Create a c# `UActorComponent` class
2- Start Editor
3- Add component to some blueprint
4- Add some `UProperty` with `EditAnywhere` flag.

Not only new variables aren't visible, existing ones don't update aswell. Like adding a `UIMin("0")` won't be reflected until editor restart.

This issue doesn't happen in all components. I traced it down to UnrealSharpCompiler.cpp:59 where some component hot reloads have `ManagedClassesToCompile` via `FCSReferencesCollection`, some don't.

CanGoktas-Nar:
During hot reload, code only checks for `ManagedClassesToCompile` but there is scenarios where there is only `ManagedComponentsToCompile` and compile gets skipped. This pr aims to fix it.

Resolves https://github.com/UnrealSharp/UnrealSharp/issues/706

### DIFF
--- a/Source/UnrealSharpCompiler/Private/UnrealSharpCompiler.cpp
+++ b/Source/UnrealSharpCompiler/Private/UnrealSharpCompiler.cpp
@@ -56,7 +56,7 @@ void FUnrealSharpCompilerModule::RecompileAndReinstanceBlueprints()
 		return;
 	}
 	
-	if (ManagedClassesToCompile.IsEmpty())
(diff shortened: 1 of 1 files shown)


FailMelon:
Thank you for the fix!

Issue #705, opened 2026-07-17
Build error when trying to install UnrealSharp in a fresh UE 5.8 project


hallgaeuer:
Steps to reproduce:

- Create new fresh UE 5.8 project (C++ based)
- `git clone https://github.com/UnrealSharp/UnrealSharp.git` into Plugins/ dir of new projects
- Try to build (I'm using Rider)

Results in the following error:
```
CONSOLE: C:\Program Files\Epic Games\UE_5.8\Engine\Build\BatchFiles\Build.bat  UnrealSharpTestEditor Win64 DebugGame -Project="D:\Projekte\Unreal\UnrealSharpTest\UnrealSharpTest.uproject" -WaitMutex  -FromMSBuild
0>Using bundled DotNet SDK version: 10.0 win-x64
0>Running UnrealBuildTool: dotnet "..\..\Engine\Binaries\DotNET\UnrealBuildTool\UnrealBuildTool.dll" UnrealSharpTestEditor Win64 DebugGame -Project="D:\Projekte\Unreal\UnrealSharpTest\UnrealSharpTest.uproject" -WaitMutex  -FromMSBuild
0>Log file: C:\Users\hallg\AppData\Local\UnrealBuildTool\Log.txt
0>Determining max actions to execute in parallel (16 physical cores, 32 logical cores)
0>  Executing up to 16 processes, one per physical core
0>Creating makefile for UnrealSharpTestEditor (no existing makefile)
0>UbaServer - Listening on 0.0.0.0:1345
0>Skipping net8.0 project D:\Projekte\Unreal\UnrealSharpTest\Plugins\UnrealSharp\Source\UnrealSharpManagedGlue\UnrealSharpManagedGlue.ubtplugin.csproj
0>Unhandled exception: FileNotFoundException: Could not load file or assembly 'D:\Projekte\Unreal\UnrealSharpTest\Plugins\UnrealSharp\Source\UnrealSharpManagedGlue\bin\Development\win-x64\UnrealSharpManagedGlue.dll'. The system cannot find the file specified.
0>   at System.Reflection.Assembly.LoadFile(String path)
0>   at EpicGames.UHT.Utils.UhtTables.LoadAssembly(String assemblyFilePath) in D:\build\++UE5\Sync\Engine\Source\Programs\Shared\EpicGames.UHT\Utils\UhtTables.cs:line 577
[...]

PR #710, opened 2026-07-26
Fix Unreal Engine 5.8 laucnher compatibility

APinchofDill:
**Cause**: launcher installed UE 5.8 does not  include the generated UE5Rules.csproj UnrealSharp expected, so engine version symbols were empty and FInstancedStruct went into the wrong C# namespace.

**Fix**: fallback detection in UnrealSharp.Automation.props and preprocessorexporter.cs to correctly select net10 in the launcher version.

discord ticket: [here](https://discord.com/channels/1153570806417342466/1530720394531700756)

### DIFF
--- a/Source/UnrealSharpManagedGlue/Exporters/PreprocessorExporter.cs
+++ b/Source/UnrealSharpManagedGlue/Exporters/PreprocessorExporter.cs
@@ -2,7 +2,7 @@
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text.RegularExpressions;
+using System.Text.Json;
 using System.Xml.Linq;
 using UnrealSharpManagedGlue.SourceGeneration;
 using UnrealSharpManagedGlue.Utilities;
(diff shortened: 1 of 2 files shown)


hallgaeuer:
This fixes my issue #705. Thanks!

OlssonDev:
` UE5Rules.csproj` exists for me in 5.8. Doesn't it for you?

FailMelon:
> ` UE5Rules.csproj` exists for me in 5.8. Doesn't it for you?

Since UBT does not require it to be generated to compile, i guess there can be certain circumstances that this project does not exist but the project bindings do. I'm not exactly sure what this circumstances are but because of this i think i'd recommend not using the csproj as a source of truth for the rules and instead replace it entirely with this logic. 

The logic posted in it is similar logic unreal uses to generate the UE5Rules.csproj so it should be the exact same the only preprocessors missing would be these 2

```CSharp
PreprocessorDefines.Add("WITH_FORWARDED_MODULE_RULES_CTOR");
PreprocessorDefines.Add("WITH_FORWARDED_TARGET_RULES_CTOR");
```

This is the logic from `Engine\Source\Programs\UnrealBuildTool\Configuration\Rules\RulesAssembly.cs`
```CSharp
		/// <summary>
		/// Finds all the preprocessor definitions that need to be set for the current engine.
		/// </summary>
		/// <returns>List of preprocessor definitions that should be set</returns>
		public static List<string> GetPreprocessorDefinitions()
		{
			List<string> PreprocessorDefines = new List<string>();
			PreprocessorDefines.Add("WITH_FORWARDED_MODULE_RULES_CTOR");
			PreprocessorDefines.Add("WITH_FORWARDED_TARGET_RULES_CTOR");

			// Define macros for the Unreal engine version, starting with 4.17
			// Assumes the current MajorVersion is 5
			BuildVersion? Version;
			if (BuildVersion.TryRead(BuildVersion.GetDefaultFileName(), out Version))
			{
				for (int MinorVersion = 17; MinorVersion <= 30; MinorVersion++)
				{
					PreprocessorDefines.Add(String.Format("UE_4_{0}_OR_LATER", MinorVersion));
				}
				for (int MinorVersion = 0; MinorVersion <= Version.MinorVersion; MinorVersion++)
				{
					PreprocessorDefines.Add(String.Format("UE_5_{0}_OR_LATER", MinorVersion));
				}
			}
			return PreprocessorDefines;
		}
```

`BuildVersion.GetDefaultFileName()` with Returning Build.version json file in the engines build directory.

OlssonDev:
I fixed this on latest with @FailMelon findings. Thank you for the contribution ❤️

PR #721, opened 2026-09-01
Use dotnet executable in selected sdk

hez2010:
Otherwise it can collide with newer SDKs installed on the PC, causing the build to fail.


OlssonDev:
Thank you for the fix!

PR #722, opened 2026-09-12
Fix C# hot reload default propagation

FailMelon:
This PR fixes two issues with cdo defaults.

- Fixed issue where defaults would overwrite the user set values, this was causing values to flip flop.
- Fixed an issue where C# constructor body changes were not getting detected.
- Fixed incremental hot reload baseline tracking by seeding cached syntax trees from the active compilation and preserving recently removed trees, preventing remove/add file watcher events from causing new C# syntax to be compared against itself.
- Recover struct memory after hot reload (this is done to preserve user unsaved data.)

### DIFF
--- a/Managed/UnrealSharp/UnrealSharp.Editor/IncrementalCompilationManager.cs
+++ b/Managed/UnrealSharp/UnrealSharp.Editor/IncrementalCompilationManager.cs
@@ -20,6 +20,7 @@ public static class IncrementalCompilationManager
                || state.Driver == null
                || state.InitialCompilation == null
                || state.TreesByPath == null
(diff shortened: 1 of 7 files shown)


OlssonDev:
Feels like this solution will grow exponentially very fast since we're capturing every object in an assembly on every hot reload

FailMelon:
> Feels like this solution will grow exponentially very fast since we're capturing every object in an assembly on every hot reload

Yeah i thought about going serialize from disk route but the problem with that is any unsaved data in the editor will get overwritten during a hot reload, i'm not sure how to get unsaved data if possible.

PR #723, opened 2026-09-12
Fixed Stale SCS attachment data surviving hot reload.

FailMelon:
This PR Fixes Stale SCS attachment data which could survive hot reload, causing components to appear under the wrong parent until the editor rebuilt the hiearchy from scratch via a restart.


OlssonDev:
Thank you for the fix!

PR #725, opened 2026-09-19
Add FVector2f support

hallgaeuer:
Adds simple support for FVector2f. 
I added some ctors and implicit conversions between FVector2D, but no helper methods or math operators. (could add them if requested).

Why even use FVector2f instead of FVector2D:
I'm trying to integrate SlateIM into UnrealSharp (it actually already exposes some APIs to blueprints / U#). 
The FSlateIMWindowParams struct unfortunately only takes an FVector2f, requiring the U# support of this struct.


halo11009:
Maybe mapping FVector2f to FVector2D in C# through Config/Default.UnrealSharpTypes.json could simplify things by removing the need for multiple 2D vector structs.

OlssonDev:
> Maybe mapping FVector2f to FVector2D in C# through Config/Default.UnrealSharpTypes.json could simplify things by removing the need for multiple 2D vector structs.

FVector2D are doubles and FVector2f are floats

OlssonDev:
Thank you for the contribution!

